Issue #24, opened 2017-09-05
Downloads Episode Selection


sandikbarr:
User can select which episodes are to be charted. Options include Last 5 Released, Last 10 Released, All Others/All Episodes. When some episodes are selected, the option is "All Others"; when no other episodes are selected, the option is "All Episodes". All Episodes will chart at the podcast level, not the individual episode level.

Issue #37, opened 2017-10-02
canned date range selection overridden when match more than one option


sandikbarr:
When the canned date range matches more than one option, the selection can be overridden by the auto setSelectedIfFilterIsRange. For example, this month of October 2017 the 1st is on a Sunday, so for this first week of the month 'This week' also matches 'This month'. When the user selects 'This week', it is overridden by 'This month'. There is also a test that has been disabled due to this bug: 'should go to prev range when PREV button is click' that should be re-enabled when fixed.

PR #38, opened 2017-10-02
adds episode selection to downloads

sandikbarr:
#24


src/app/shared/filter/canned-range.component.spec.ts
```diff
@@ -86,7 +86,7 @@ describe('CannedRangeComponent', () => {
     expect(comp.filter.endDate.valueOf()).toEqual(comp.endOfTodayUTC().valueOf());
   });
 
-  it('should go to prev range when PREV button is click', () => {
+  xit('should go to prev range when PREV button is click', () => {
```
sandikbarr: would be a good week to address this one: https://github.com/PRX/metrics.prx.org/issues/37

src/app/shared/filter/canned-range.component.ts
```diff
@@ -101,7 +101,7 @@ export class CannedRangeComponent implements OnInit, OnDestroy {
     this.whenOptions.push([LAST_MONTH, {beginDate: utcBeginDate.toDate(), endDate: utcEndDate.toDate()}]);
 
     // first of this month - 1 day and 2 months
-    utcEndDate = this.endOfTodayUTC().date(1).subtract(1, 'days').subtract(2, 'months');
+    utcEndDate = this.endOfTodayUTC().date(1).subtract(2, 'months').subtract(1, 'days');
```
sandikbarr: Prior 3 months was ending on 7/30. Oops.

src/app/shared/util/metrics.util.ts
```diff
@@ -0,0 +1,57 @@
+import { PodcastModel, EpisodeModel, FilterModel, MetricsType, PodcastMetricsModel, EpisodeMetricsModel } from '../../ngrx/model';
+
```
sandikbarr: I separated these filtering and utility functions from the selectors for better organization.

src/app/shared/filter/episodes.component.ts
```diff
@@ -0,0 +1,48 @@
...
+import { selectEpisodeFilter, selectEpisodes } from '../../ngrx/reducers';
+import { EpisodeModel } from '../../ngrx/model';
+import { CastleFilterAction } from '../../ngrx/actions';
+
+@Component({
+  selector: 'metrics-episodes',
+  template: `
+    <span>Episodes:</span>
+    <span>
+      <prx-select [options]="allEpisodeOptions" [selected]="selected" searchable="true" (onSelect)="onEpisodesChange($event)"></prx-select>
```
cavis: Seems like there are some UX issues with how the select works here.  May need to address them in styleguide at some point.

No results rendered until you start typing, so it's hard to see/alter the selections.  In this case, would probably be best if it rendered the "most recent N", but typing in the filter can get you to the older episodes.  Not sure what to do about the "selected but not visible per the filter" problem.

sandikbarr:
@cavis I'm not clear on what you're asking me to change. I can add those canned options (not sure they were really shown to be helpful, given the last episode overwhelming the chart and producers only really looking at the most recent episode), but they're not going to solve the problem of options not being visible in `prx-select`.

cavis:
No changes!  I just tend to use "request changes" to explicitly block merging while i ask questions.

Hashed out in slack - will merge this functionality, and open some other tickets to improve the upstream prx-select.

PR #39, opened 2017-10-04
dockerize

sandikbarr:
dockerize metrics for CICD


Dockerfile
```diff
@@ -0,0 +1,39 @@
...
+ENTRYPOINT [ "/tini", "--", "./bin/application" ]
+CMD [ "serve" ]
+
+ADD ./package.json ./
+ADD ./yarn.lock ./
+RUN apk --update add curl git && \
+  curl -Ls "https://github.com/dustinblackman/phantomized/releases/download/2.1.1/dockerized-phantomjs.tar.gz" | tar xz -C / && \
+  npm install yarn --global && \
+  yarn install --no-progress --silent && \
+  apk del curl && \
```
kookster: I see publish has a `npm cache clean && \` after this line, is that not a good idea? should it be removed from publish too?

docker-compose-ci.yml
```diff
@@ -0,0 +1,5 @@
+metrics:
```
kookster: Is this file being used?

The `.prxci` file doesn't specify it, i.e. it would need to be:
```
docker-compose -f docker-compose-ci.yml build
```

so I think the CICD build is just using the regular default `docker-compose.yml`, and you can delete this file.

docker-compose.yml
```diff
@@ -0,0 +1,23 @@
...
+    - ./package.json:/app/package.json
+    - ./protractor.conf.js:/app/protractor.conf.js
+    - ./tsconfig.json:/app/tsconfig.json
+    - ./tslint.json:/app/tslint.json
+    - ./.git:/app/.git
+  env_file:
+    - .env
+  ports:
+    - "4202:4202"
+    - "4203:4203"
```
kookster: so nice the ports don't collide with publish

Dockerfile
sandikbarr: It's not a thing anymore. It wasn't causing problems in Publish yesterday, but it should be removed.
```
$ npm cache clean
npm ERR! As of npm@5, the npm cache self-heals from corruption issues and data extracted from the cache is guaranteed to be valid. If you want to make sure everything is consistent, use 'npm cache verify' instead.
npm ERR! 
npm ERR! If you're sure you want to delete the entire cache, rerun this command with --force.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/sandibarr/.npm/_logs/2017-10-04T15_01_28_056Z-debug.log
```

docker-compose-ci.yml
sandikbarr: You're probably right, but the guide says I should have both docker-compose.yml and docker-compose-ci.yml https://github.com/PRX/internal/pull/108/files#diff-5b3ca927271ab217a5d8c663b572f159R18

sandikbarr:
@kookster removed docker-compose-ci.yml, back to you

Dockerfile
kookster: I had no idea, thanks!

docker-compose-ci.yml
kookster: Yeah, it's optional, for cases where you need the ci to do something slightly differently.
For example, we don't start up the worker for testing cms or feeder, so those get left out for ci.

PR #43, opened 2017-10-06
users need to be able to select which podcast

sandikbarr:
I went ahead and made a decision about how to select a podcast. If it's not what we want, it's just software, we can change it.

Branched from fix/fixes (in PR) because I needed one of those fixes to get this working.


src/app/shared/filter/podcasts.component.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+      comp.store.dispatch(new CastleFilterAction({filter}));
+      comp.store.dispatch(new CmsPodcastFeedAction({podcast: podcasts[0]}));
+    });
+  }));
+
+  it('should initialize selected podcast according to filter', () => {
+    expect(comp.selectedPodcast.seriesId).toEqual(podcasts[0].seriesId);
+  });
+
+  it('should only show podcast title rather than dropdown if there is only one podcast', () => {
```
gcampo88: nitpickiest of nitpicky comments: i would nix 'only' from the description since this doesn't check the case where there's more than one podcast. (i hate myself for how nitpicky that is)

src/app/shared/filter/podcasts.component.ts
```diff
@@ -0,0 +1,45 @@
...
+export class PodcastsComponent implements OnInit, OnDestroy {
+  podcastFilterSubscription: Subscription;
+  podcastsSubscription: Subscription;
+  selectedPodcast: PodcastModel;
+  allPodcastOptions = [];
+
+  constructor(public store: Store<any>) {}
+
+  ngOnInit() {
+    this.podcastFilterSubscription = this.store.select(selectPodcastFilter).subscribe((filteredPodcast: PodcastModel) => {
```
gcampo88: oooh, I'm a little late to the party but this createSelector business is awesome!

src/app/app.component.html
```diff
@@ -1,7 +1,6 @@
 <prx-auth [host]="authHost" [client]="authClient"></prx-auth>
 <prx-header>
-  <prx-navitem *ngIf="loggedIn" route="/" text="Home"></prx-navitem>
-  <prx-navitem *ngIf="loggedIn" route="/search" text="Search"></prx-navitem>
+  <metrics-podcasts *ngIf="loggedIn"></metrics-podcasts>
```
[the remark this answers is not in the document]
sandikbarr: yeah... there are no routes at this point, so we're always on home. The PRX logo also goes to the root, so Home would be redundant if there were routes.

src/app/shared/filter/podcasts.component.ts
sandikbarr: immutability rules still apply, but can be handy when you want a smaller/deeper slice of state

src/app/shared/filter/podcasts.component.spec.ts
sandikbarr: Don't hate yourself! Tests should also serve to document the application, and if the description isn't working, the documentation isn't so good.
More than one podcast case test is below.

Issue #46, opened 2017-10-12
add apply button to make changes to parameters all at once


sandikbarr:
Load time are not instantaneous, so the user should be able to make multiple changes to the parameters before loading new data. 
* [x] The Next and Prev buttons will not be dependent on the apply button and can be clicked independently
* [x] The Next and Prev buttons will be moved to surround the Apply action button
* [x] The podcast dropdown is a separate control and will be applied on change
* [x] Left hand drop down controls will be left aligned
* [x] When the apply button is clicked, ~all data should be cleared~ data not included in the new filter should not be shown

Issue #26, opened 2017-09-05
Downloads Table View


sandikbarr:
Provide table view of downloads data. To reach feature parity with podtrac, we need to show all of the episodes in the table view, not just the last 5 released or just what is currently in the chart. Performance still needs to be proven on whether we should provide this in a default view, so perhaps this feature can be toggled on.

sandikbarr:
upped estimate to 2 points because I'll need to move the episode sorting from within the chart component to application state

sandikbarr:
Thought we might try to get @gcampo88 involved on the front end here.

PR #50, opened 2017-10-12
Episode downloads table 

gcampo88:
#26. Notes:
- Per team discussion on slack, table just shows selected episodes, and we will open tix on both metrics and castle to get it to request + show data for all episodes. 
- Existing styling funkiness -- seemed like `.controls` and `.content` were meant to take up 20% and 80% of the available horizontal space, respectively, but the controls were overflowing their container and pushing everything off the right. I've now given both a `max-width` of their respective percentages, which is making the controls section look a little cramped. Since it sounds like there [may be a redesign ](https://github.com/PRX/metrics.prx.org/issues/46)of that section imminently, okay to leave this cramped-ness for now for addressing in the later ticket redesigning the controls section?


src/app/downloads/downloads-table.component.ts
```diff
@@ -0,0 +1,135 @@
...
+        <tr>
+          <th>Episode</th>
+          <th>Release Date</th>
+          <th>Total for period</th>
+          <th *ngFor="let date of dateRange">{{date}}</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr *ngFor="let episode of episodeTableData">
+          <td>{{episode.title}}</td>
```
sandikbarr: Nice scrolling action. Any chance we can make this part of the table "sticky"? Andrew suggested we make these headings always visible when the table is horizontally long.

src/app/downloads/downloads-table.component.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    expect(comp.episodeTableData[0].totalForPeriod).not.toBeNull();
+  });
+
+  it('should sort episodes by published date', () => {
+    expect(comp.episodeTableData[0].title).toEqual('A More Recent Pet Talk Episode');
+  });
+
+  it('should not show up without data to display', () => {
+    comp.episodeTableData = [];
+    expect(de.query(By.css('metrics-downloads-table'))).toBeNull();
```
sandikbarr: Should this look for 'table' here instead of 'metrics-downloads-table'?

src/app/downloads/downloads-table.component.spec.ts
```diff
@@ -0,0 +1,142 @@
...
+    const differentPodcast: PodcastModel = {
+      doc: undefined,
+      seriesId: 37801,
+      title: 'Totally Not Pet Talks Daily'
+    };
+    comp.store.dispatch(new CastleFilterAction({filter: {podcast: differentPodcast}}));
+    expect(comp.episodeTableData.length).toEqual(0);
+  });
+
+  it('should display episode total downloads for period', () => {
```
sandikbarr: I feel like this test is just testing the getTotal utility function. Is there something that would cause it to not display the total here that should be tested?

src/app/downloads/downloads-table.component.ts
```diff
@@ -0,0 +1,136 @@
...
+      <thead>
+        <tr>
+          <th class="sticky">Episode</th>
+          <th>Release Date</th>
+          <th>Total for period</th>
+          <th *ngFor="let date of dateRange">{{date}}</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr *ngFor="let episode of episodeTableData">
```
sandikbarr: The first table row below the date headers should be a row of podcast metrics of "All Episodes" https://projects.invisionapp.com/share/X6CQZU4MQ#/screens/249402553

src/app/downloads/downloads-table.component.ts
```diff
@@ -0,0 +1,136 @@
...
+    this.episodeMetricsStoreSub = this.store.select(selectEpisodeMetrics).subscribe((episodeMetrics: EpisodeMetricsModel[]) => {
+      this.episodeMetrics = filterEpisodeMetrics(this.filter, episodeMetrics, 'downloads');
+      this.buildEpisodeMetrics();
+    });
+  }
+
+  buildEpisodeMetrics() {
+    if (this.episodes && this.episodeMetrics && this.episodeMetrics.length) {
+      this.episodeTableData = this.episodeMetrics
+        .filter((epMetric) => !!this.episodes.find(ep => ep.id === epMetric.id))
```
sandikbarr: I'm not sure what this filter is for. episodes/allPodcastEpisodes are filtered by the selected podcast, episode metrics are also already filtered by the selected podcast. If anything, it seems more likely that episode metrics would not contain episodes that are in episodes/allPodcastEpisode.

src/app/downloads/downloads-table.component.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (this.episodeTableData && this.episodeTableData[0] && this.episodeTableData[0].downloads) {
+      this.dateRange = this.episodeTableData[0].downloads.map(d => this.dateFormat(new Date(d.date)));
+    }
+  }
+
+  dateFormat(date: Date): string {
+    if (this.filter && this.filter.interval) {
+      switch (this.filter.interval.key) {
+        case INTERVAL_DAILY.key:
+          return dayMonthDate(date);
```
sandikbarr: Users may want that day of the week date format on the table too (a producer requested it), but we can wait for user feedback on that.

src/app/downloads/downloads.component.ts
```diff
@@ -139,7 +140,7 @@ export class DownloadsComponent implements OnInit, OnDestroy {
           this.error = 'An error occurred while requesting podcast metrics on ' + this.filter.podcast.title;
           console.error(err);
         } else {
-          this.error = this.filter.podcast.title + ' podcast has no download metrics.';
+          this.error = this.filter.podcast.title + ' podcast has no download metrics for some or all of this time.';
```
sandikbarr: This change to the error message seems inaccurate, but the first attempt at providing an error message wasn't accurate either. If Castle doesn't have metrics for all or part of the time, it will populate the rest of the dates with zeroes (this was not initially true.) When I'm seeing these it's other errors occurring in Castle (that should be investigated.) I've also noticed that half assed attempt at showing these errors only supports one error at a time. Ticket?

src/app/downloads/downloads-table.component.css
```diff
@@ -0,0 +1,49 @@
+:host {
+  width: 100%;
```
sandikbarr: do both :host and table need width and overflow-x? Is it possible for the table to be relative instead? (I can see wanting to put the Export to CSV link in here above the table.)

src/app/downloads/downloads.component.css
```diff
@@ -5,6 +5,7 @@
 }
 .controls {
   flex: 1 1 20%;
+  max-width: 20%;
```
sandikbarr: Can we change this to 30/70 so it doesn't look so terrible in the meantime?
`.controls {
  flex: 1 1 30%;
  max-width: 30%;
}
.content {
  flex: 7 7 70%;
  max-width: 70%;
}`

src/app/downloads/downloads.component.ts
gcampo88: ticket sounds good to me -- will nix the changed error message for now. (in dev, i was getting 'has no download metrics' for stuff that did have data for part of the time, going off latest castle)

src/app/downloads/downloads-table.component.spec.ts
gcampo88: Good point. I can just check for the existence of `totalForPeriod` on each row -- what I wanted to have in the test was something that checks for the totaling having happened for every row.

src/app/downloads/downloads.component.ts
sandikbarr: Opened a ticket: https://github.com/PRX/metrics.prx.org/issues/53

src/app/downloads/downloads-table.component.ts
gcampo88: good idea!

gcampo88:
okey doke @sandikbarr think it's ready for you again

Issue #53, opened 2017-10-16
handle errors from Castle


sandikbarr:
Occasionally metrics gets an error from Castle. It's currently displaying "<title> [episode|podcast] has no download metrics". However, that should not be the case, Castle should return data with zero valued entries for the dates requested. (In testing, it looks like Castle may be returning errors when it has data for only part of the date range requested.)
The error message in the UI should be
* [x] accurate for the error (if it's just "error occurred requesting metrics" that would be more accurate than saying it doesn't have download metrics because that's likely not what is happening)
* [x] display all errors rather than just the last one

sandikbarr:
Still need to investigate these errors and open castle tickets, but now they're handled better in metrics at least.

Issue #28, opened 2017-09-13
Discuss feature set on mobile


farski:
I think we know there is demand for good mobile support, but the marginal cost of work as we approach 100% feature parity with desktop potentially gets fairly high. Worth having a think through if that's true at all, or what we'd do if it is.

pkarman:
Not having mobile support as part of the "beta" feature set seems like a blocker to me. Is there a strong reason for not using a responsive design so that things work on smaller screen sizes? Seems like if we know folks are going to use their pads and phones, we wouldn't want to promote the product if it does work on those devices.

pkarman:
@farski @kookster @sandikbarr and @pkarman met to discuss. We agreed that responsive design should take mobile into account for sidebar (#130) and that we're aiming for full feature parity regardless of form factor.

We'll wait till after beta release to gather feedback on specific interactions that require some design attention on mobile devices.

PR #55, opened 2017-10-20
adds apply button

sandikbarr:
#46 Load time are not instantaneous, so the user should be able to make multiple changes to the parameters before loading new data.

* The Next and Prev buttons will not be dependent on the apply button and can be clicked independently
* The Next and Prev buttons will be moved to surround the Apply action button
* The podcast dropdown is a separate control and will be applied on change
* Left hand drop down controls will be left aligned
* When the apply button is clicked, ~all data should be cleared~ data not included in the new filter should not be shown <-- I changed this requirement somewhat during development. Clearing the table and chart data was fighting against the incoming stream of state changes. It doesn't show the old data while loading anymore, and I hope that's good enough.

I have also incorporated some feedback from the internal review into this PR. I haven't done the routing or the move to feeder yet, but I'm hoping that we can do another quick round of review before we start user testing again.
* Users are told they don't have any podcasts rather than looking at an infinite spinner
* Don't show the filter controls on initial load of the app
* #53 Better error handling from Castle (but I still don't know why it's throwing those random 500s)


src/app/downloads/downloads-chart.component.ts
```diff
@@ -22,74 +22,68 @@ export class DownloadsChartComponent implements OnDestroy {
...
-      this.colors = generateShades(this.episodeMetrics.length);
-      this.episodeChartData = this.episodeMetrics
-        .sort((a: EpisodeMetricsModel, b: EpisodeMetricsModel) => {
-          return getTotal(metricsData(filter, b, 'downloads')) - getTotal(metricsData(filter, a, 'downloads'));
-        })
-        .map((metrics: EpisodeMetricsModel, i) => {
-          const episode = filter.episodes.find(ep => ep.id === metrics.id);
-          return this.mapEpisodeData(episode, metricsData(filter, metrics, 'downloads'), this.colors[i]);
-        });
+  applyFilterToExistingData() {
```
gcampo88: this is nice!

sandikbarr: thanks, it is definitely getting better! 😸

src/app/downloads/downloads.component.css
```diff
@@ -1,27 +1,45 @@
...
+    max-width: 60%
+  }
 }
-.bar {
-  width: 75%;
-  padding: 0 40px;
-  margin: 20px 0px;
-  border: 1px solid #a3a3a3;
+@media only screen
+and (max-device-width: 568px) {
```
gcampo88: responsive 🌮 ! 
left-align looks slightly odd with nothing to its right -- could center the controls in their container in this rule? (or can wait for user testing, i don't trust my own opinions) 

image

sandikbarr: I have some different ideas about the mobile layout too like hiding controls in a menu and when open putting date controls to the left and interval and episodes on the right (if that would fit.) I'm imagining that we'll actually get some mobile specific tickets out of this meeting when we have it https://github.com/PRX/metrics.prx.org/issues/28. I just added this bit of media queries because Peter was looking at the app on smaller screen sizes already and commented about it, so this should just make it less bad.

src/app/ngrx/reducers/filter.reducer.ts
```diff
@@ -20,6 +20,13 @@ export function FilterReducer(state: FilterModel = initialState, action: ActionW
         // but we can and maybe should do it in action creators instead of components)
         newState.when = action.payload.filter.when;
       }
+      // the reason that there is a range separate from the standard range "when"
+      //  is because the when value becomes undefined when the begin and end dates do not
+      //  match one of the valid options; however, the range sticks around so that prev/next
+      //  buttons can still be used to page through based on the last selected range and the date pickers
```
gcampo88: 👍  v helpful comment

gcampo88: (could be a reason to rename 'when' to something more specific to canned-option nature?)

sandikbarr: Heh, I feel like this is a bit of a surprise feature but still maybe better than disabling the prev and next when it's not a standard range.

sandikbarr: yeah, maybe "standardRange" instead of "when"? Let's see if this standard vs custom idea sticks.

src/app/shared/filter/date/date-range.component.ts
```diff
@@ -0,0 +1,30 @@
...
+  `
+})
+export class DateRangeComponent {
+  @Input() filter: FilterModel;
+  @Output() dateRangeChange = new EventEmitter<DateRangeModel>();
+
+  onDateRangeChange(dateRange: DateRangeModel) {
+    const when = getWhenForRange(dateRange);
+    const range = getRange(when);
+    this.dateRangeChange.emit({...dateRange, when, range});
```
gcampo88: Feels slightly odd to have two different "levels" of event emitters both called `dateRangeChange`? Will the dateRangeChange emitted by the custom date range propagate up to the metrics-filter component as well?

sandikbarr: only this one gets up to the metrics-filter, but I can rename the lower one to customDateRangeChange if that makes it less confusing

src/app/shared/filter/date/prev-date-range.component.ts
```diff
@@ -0,0 +1,112 @@
...
+
+  get prevDisabled(): string {
+    if (!this.filter || !this.filter.range) {
+      return 'disabled';
+    } else {
+      return null;
+    }
+  }
+
+  prev() {
```
gcampo88: was going to ask why the logic of getting the new date filters in this component doesn't look more like that of the next-date-range component -- i assume it's because navigating to previous time periods is a little trickier since we might not be in a complete one now? (e.g. if we're on "this week" but it's only 4 days into the week, couldn't navigate back by the same amount of days and expect it to say "last week") ?

src/app/shared/filter/filter.component.ts
```diff
@@ -0,0 +1,57 @@
...
+@Component({
+  selector: 'metrics-filter',
+  template: `
+    <metrics-date-range [filter]="filter" (dateRangeChange)="onDateRangeChange($event)"></metrics-date-range>
+    <hr>
+    <metrics-interval [filter]="filter" (intervalChange)="onIntervalChange($event)"></metrics-interval>
+    <metrics-episodes [filter]="filter" (episodesChange)="onEpisodesChange($event)"></metrics-episodes>
+    <div class="buttons">
+      <metrics-prev-date-range></metrics-prev-date-range>
+      <button (click)="onApply()">Apply</button>
```
gcampo88: Could be nice to disable this when filter hasn't been changed -- visual confirmation to user that what they're looking at matches what's in controls

src/app/shared/filter/date/prev-date-range.component.ts
sandikbarr: that's exactly it, it's for dealing with those partial ones for the standard ranges in the present

src/app/shared/util/metrics.util.ts
```diff
@@ -22,9 +24,34 @@ export const filterEpisodes = (filter: FilterModel, episodes: EpisodeModel[]) =>
...
+  const begin = findEntryByDate(beginDate);
+  const end = findEntryByDate(endDate);
+  if (begin !== -1 && end !== -1) {
+    return metrics.slice(begin, end + 1);
+  } else {
+    return null; // no partial data
+  }
+};
+
+export const findPodcastMetrics = /* TODO: findPodcastMetrics? */
```
gcampo88: this line seems like it doubts itself

src/app/shared/util/metrics.util.ts
```diff
@@ -33,13 +60,22 @@ export const filterPodcastMetrics = (filter: FilterModel, podcastMetrics: Podcas
...
   (filter: FilterModel, episodeMetrics: EpisodeMetricsModel[], metricsType: MetricsType): EpisodeMetricsModel[] => {
-    if (filter.podcast && filter.episodes && filter.interval && episodeMetrics) {
-      const metricsProperty = filter.interval.key + metricsType.charAt(0).toUpperCase() + metricsType.slice(1);
-      return episodeMetrics.filter(metric => metric.seriesId === filter.podcast.seriesId &&
-      filter.episodes.map(e => e.id).indexOf(metric.id) !== -1 &&
-      metric[metricsProperty]);
-    }
-  };
+  if (filter.podcast && filter.episodes && filter.interval && filter.beginDate && filter.endDate && episodeMetrics) {
+    const metricsProperty = filter.interval.key + metricsType.charAt(0).toUpperCase() + metricsType.slice(1);
```
gcampo88: could lift `filter.interval.key + metricsType.charAt(0).toUpperCase() + metricsType.slice(1)` into a small helper? since it's used in each of these. /shrug

src/app/ngrx/reducers/filter.reducer.ts
gcampo88: that sounds good!

src/app/downloads/downloads.component.css
gcampo88: 👍

src/app/shared/filter/date/date-range.component.ts
gcampo88: i think that could help!

src/app/shared/filter/date/prev-date-range.component.ts
gcampo88: 👍

sandikbarr:
Thanks @gcampo88. I'll go ahead and make those changes before merging.

src/app/shared/util/metrics.util.ts
sandikbarr: oh, I did end up renaming that function and didnt remove the comment about it

sandikbarr:
now I feel like I might want to rename the filter.range to prevNextRange. This is why they say software is never finished... There is more refactoring to come with the move to feeder, so maybe in the next round. Back to you @gcampo88!

src/app/shared/filter/date/date-range.component.spec.ts
```diff
@@ -43,17 +43,17 @@ describe('DateRangeComponent', () => {
...
       endDate: endOfLastMonthUTC().toDate()
     };
-    comp.onDateRangeChange(range);
-    const inSync = {...range, when: LAST_MONTH, range: getRange(LAST_MONTH)};
+    comp.onCustomRangeChange(range);
+    const inSync = {...range, standardRange: LAST_MONTH, range: getRange(LAST_MONTH)};
     expect(comp.dateRangeChange.emit).toHaveBeenCalledWith(inSync);
   });
 
   it('keeps custom range dates in sync with standard range "when" and prev/next range', () => {
```
gcampo88: two lingering references to "when" in the test descriptions in this file

src/app/shared/filter/date/date-range.component.ts
```diff
@@ -1,30 +1,30 @@
...
   selector: 'metrics-date-range',
   template: `
-    <metrics-standard-date-range *ngIf="filter" [interval]="filter.interval" [when]="filter.when"
-                                 (whenChange)="onWhenChange($event)"></metrics-standard-date-range>
+    <metrics-standard-date-range *ngIf="filter" [interval]="filter.interval" [standardRange]="filter.standardRange"
+                                 (standardRangeChange)="onStandardRangeChange($event)"></metrics-standard-date-range>
     <metrics-custom-date-range *ngIf="filter" [interval]="filter.interval"
                                [beginDate]="filter.beginDate" [endDate]="filter.endDate"
-                               (dateRangeChange)="onDateRangeChange($event)"></metrics-custom-date-range>
+                               (customRangeChange)="onCustomRangeChange($event)"></metrics-custom-date-range>
```
gcampo88: i like the super explicit event emitters! 👍

gcampo88:
💯

Issue #49, opened 2017-10-12
Google Analytics Event date range tracking


sandikbarr:
We have some unresolved questions about how often users will want to use the standard date ranges vs custom dates and whether they will want to adjust the time from beginning of day through end of day and which options are most used. Add Google Analytics event tracking to the date and time range controls so that different dimensions can be viewed.

PR #58, opened 2017-10-24
google analytics date range events

sandikbarr:
#49 to track the way users are using the date range and time picker controls

### DIFF
--- a/src/app/shared/filter/date/custom-date-range.component.ts
+++ b/src/app/shared/filter/date/custom-date-range.component.ts
@@ -1,6 +1,6 @@
 import { Component, Input, Output, EventEmitter } from '@angular/core';
 import { Angulartics2 } from 'angulartics2';
-import { DateRangeModel, IntervalModel, INTERVAL_HOURLY, INTERVAL_15MIN } from '../../../ngrx/model';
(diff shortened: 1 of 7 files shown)


src/app/shared/filter/date/custom-date-range.component.ts
```diff
@@ -25,8 +26,16 @@ export class CustomDateRangeComponent {
...
   @Input() beginDate: Date;
   @Input() endDate: Date;
+  @Input() podcast: PodcastModel; // podcast is only here for google events tracking on the timepicker
   @Output() customRangeChange = new EventEmitter<DateRangeModel>();
 
+  constructor(public angulartics2: Angulartics2) {}
+
+  onBeginTimeChange(date: Date) {
+    this.onBeginDateChange(date);
+    this.googleAnalyticsEvent('begin-time', date.getUTCHours());
```
gcampo88: so when begin or end time changes, this event will fire and so will the `'custom-date'` from the date range component, right?

sandikbarr: yes, and I'm ok with that. There will be two events that mean two different things we're trying to track: They used the custom date control, and they used the time picker.

PR #63, opened 2017-10-25
smarter default episode filter

sandikbarr:
Makes the default episode filter more smarter by including episodes released during the selected date range and then some before the date range up to a number divisible by 5. The reason to include episodes outside the date range is so that the beginning of the chart is populated by their activity since episodes released after the begin date would have no initial activity.

In order to accomplish this is the most straightforward manner, I changed the way the episode list gets added to the state. It was previously using basic observables/subscriptions to get each episode's podcast distribution and individually adding them to the state. With episodes trickling into those state changes, it became complicated to keep track of whether the default episode filter was complete. Now it is waiting for all those podcast distributions to return and adding them all to the state at once. It seems much more fitting to what is actually occurring anyway since those distributions are embedded in the one episodes request but getting to them with the hal logic means juggling the observables.


src/app/ngrx/reducers/episode.reducer.ts
```diff
@@ -1,39 +1,38 @@
-import { ActionTypes, ActionWithPayload, CmsEpisodeGuidPayload } from '../actions';
+import { ActionTypes, ActionWithPayload, CmsEpisodeGuidsPayload, CmsAllPodcastEpisodeGuidsAction  } from '../actions';
 import { EpisodeModel } from '../model';
 
 const initialState = [];
 
 // TODO: should be able to use ActionWithPayload<All> here with the union type,
```
sandikbarr: This All type is still garbage, and I don't know why. It hasn't actually mattered yet though, and instanceof is still a work around if it ever does.

src/app/app.component.ts
```diff
@@ -104,7 +106,7 @@ export class AppComponent implements OnInit, OnDestroy {
   getSeriesPodcastDistribution(podcast: PodcastModel) {
     podcast.doc.followItems('prx:distributions').subscribe((distros: HalDoc[]) => {
       const podcasts = distros.filter((doc => doc['kind'] === 'podcast'));
-      if (podcasts && podcasts.length > 0) {
+      if (podcasts && podcasts.length > 0 && podcasts[0]['url']) {
```
cavis: Found some of these without a URL, huh?  I hope not in prod!  Haven't seen that bug for awhile now.

src/app/downloads/downloads.component.ts
```diff
@@ -34,6 +34,7 @@ export class DownloadsComponent implements OnInit, OnDestroy {
   isEpisodeLoading = true;
   isLoadingForTheFirstTime = true;
   errors: string[] = [];
+  static DONT_BREAK_CASTLE_LIMIT = 20;
```
cavis: Nice.  Who says naming isn't easy?

src/app/downloads/downloads.component.ts
```diff
@@ -126,12 +127,17 @@ export class DownloadsComponent implements OnInit, OnDestroy {
...
-        this.filter.episodes.length < 5) {
-      // init with the first five (or less) episodes
-      const episodes = this.allPodcastEpisodes.length > 5 ? this.allPodcastEpisodes.slice(0, 5) : this.allPodcastEpisodes;
-      this.store.dispatch(new CastleFilterAction({filter: {episodes}}));
-    }
+    const episodes = [];
+    this.allPodcastEpisodes.every((episode: EpisodeModel) => {
+      episodes.push(episode);
+      if ((episode.publishedAt.valueOf() < this.filter.beginDate.valueOf() &&
+          episodes.length % 5 === 0) || episodes.length === DownloadsComponent.DONT_BREAK_CASTLE_LIMIT) {
```
cavis: _(edge case)_ Does this mean if you chose a window containing no episodes (like "10 years ago"), it would still load 5 of them from Castle?

src/app/ngrx/reducers/episode.reducer.ts
```diff
@@ -1,39 +1,38 @@
...
-      const { doc, id, title, publishedAt, feederUrl, guid } = action.payload.episode;
+    case ActionTypes.CMS_ALL_PODCAST_EPISODE_GUIDS:
       const seriesId = action.payload.podcast.seriesId;
-
-      epIdx = state.findIndex(e => e.seriesId === seriesId && e.id === id);
+      epIdx = state.findIndex(e => e.seriesId === seriesId);
       if (epIdx > -1) {
-        episode = {...state[epIdx], doc, id, seriesId, title, publishedAt, feederUrl, guid};
-        newState = [...state.slice(0, epIdx), episode, ...state.slice(epIdx + 1)];
+        const existingEpisodes = state.filter(e => e.seriesId === seriesId);
```
cavis: cleaner++

src/app/app.component.ts
```diff
@@ -117,41 +119,47 @@ export class AppComponent implements OnInit, OnDestroy {
   }
 
   getEpisodes(podcast: PodcastModel) {
+    const distObsv = [];
     podcast.doc.followItems('prx:stories', {
       per: podcast.doc.count('prx:stories'),
       filters: 'v4',
       sorts: 'released_at: desc, published_at: desc'
```
cavis: Occurs to me that maybe this should just sort by published_at?  Though since the `/authorized/stories` endpoint doesn't filter to only-published, you'd still have to check for nulls.  And we've got some more CMS changes coming anyways.  Nevermind!

src/app/app.component.ts
sandikbarr: yeah... got an error on the split, didn't really investigate 🤷‍♀️

src/app/app.component.ts
sandikbarr: honestly I probably dont need this sort since it's sorting in the episode reducer. yes, change is gonna come!

src/app/downloads/downloads.component.ts
sandikbarr: hmm, yes. This is not ideal, but it is kind of an edge case to get it to load like this... I'm thinking if this default episode filter sticks, it could shift with the user selected dates. I feel like there will be more adjustments made here.
screen shot 2017-10-25 at 6 41 32 pm

Issue #60, opened 2017-10-24
show a stacked bar chart instead of a stacked area when small number of data points


sandikbarr:
If data points is less than 4, show a stacked bar chart instead of stacked area.

Issue #59, opened 2017-10-24
support weekly and monthly interals


sandikbarr:
Weekly and monthly intervals have been added to Castle, add support to metrics FE.

PR #64, opened 2017-10-26
add support for weekly and monthly intervals

sandikbarr:
#59 support weekly and monthly intervals
 * weekly and monthly are available in the interval dropdown
 * when choosing an interval, if the dates do not match up, they will be normalized to the beginning and end of the interval. For hourly and monthly data, the beginning of the interval is used because the data is inclusive and the timepicker doesn't have any steps between intervals. In order to allow the user to make adjustments to the filter parameters without taking away options, the interval is not limited to the date range selected; instead the date range is normalized on Apply

#60 this really felt incomplete without also doing the change to show a bar chart when there are less than 4 data points

### DIFF
--- a/src/app/downloads/downloads-chart.component.ts
+++ b/src/app/downloads/downloads-chart.component.ts
@@ -6,9 +6,8 @@ import { EpisodeMetricsModel, PodcastMetricsModel, EpisodeModel, FilterModel,
   INTERVAL_MONTHLY, INTERVAL_WEEKLY, INTERVAL_DAILY, INTERVAL_HOURLY, INTERVAL_15MIN } from '../ngrx/model';
 import { selectFilter, selectPodcastMetrics, selectEpisodeMetrics } from '../ngrx/reducers';
 import { findPodcastMetrics, filterEpisodeMetrics, metricsData, getTotal } from '../shared/util/metrics.util';
-import { mapMetricsToTimeseriesData, subtractTimeseriesDatasets,
-  UTCDateFormat, monthYearFormat, dayMonthDateFormat, dailyDateFormat, hourlyDateFormat,
-  neutralColor, generateShades } from '../shared/util/chart.util';
+import { mapMetricsToTimeseriesData, subtractTimeseriesDatasets, neutralColor, generateShades } from '../shared/util/chart.util';
+import { UTCDateFormat, monthYearFormat, dayMonthDateFormat, dailyDateFormat, hourlyDateFormat } from '../shared/util/date.util';
 
 @Component({
   selector: 'metrics-downloads-chart',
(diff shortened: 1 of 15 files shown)


src/app/downloads/downloads-chart.component.ts
```diff
@@ -112,12 +112,16 @@ export class DownloadsChartComponent implements OnDestroy {
   }
 
   dateFormat(): Function {
-    if (this.filter && this.filter.interval) {
-      switch (this.filter.interval.key) {
-        case INTERVAL_DAILY.key:
+    if (this.filter) {
+      switch (this.filter.interval) {
+        case INTERVAL_MONTHLY:
+          return monthYearFormat;
```
sandikbarr: maybe this should be formatted Oct '17 instead of 10/17? I imagine date formatting will come up either in internal review or user testing.

src/app/shared/util/date.util.ts
```diff
@@ -264,7 +302,55 @@ export const getAmountOfIntervals = (beginDate: Date, endDate: Date, interval: I
       return 1 + (duration / (1000 * 60 * 60));
     case INTERVAL_DAILY:
       return 1 + (duration / (1000 * 60 * 60 * 24));
+    case INTERVAL_WEEKLY:
+      return 1 + (duration / (1000 * 60 * 60 * 24 * 7));
+    case INTERVAL_MONTHLY:
+      return 1 + (12 * endDate.getUTCFullYear() - 12 * beginDate.getUTCFullYear()) + (endDate.getUTCMonth() - beginDate.getUTCMonth());
```
cavis: So these are just used by GA then?  (Attempting to figure out where/how this is used, since monthly is non-consistent one).

src/app/shared/filter/filter.component.spec.ts
```diff
@@ -85,4 +86,21 @@ describe('FilterComponent', () => {
     comp.onApply();
     expect(comp.store.dispatch).toHaveBeenCalled();
   });
+
+  it('should normalize dates to interval when interval is selected', () => {
+    comp.onIntervalChange(INTERVAL_MONTHLY);
```
cavis: I notice that selecting Weekly tends to un-select the canned date ranges (since the weeks hardly ever line up with those canned ranges, i'm assuming).  Not a huge deal though.

src/app/shared/util/date.util.ts
sandikbarr: yep, just used for the value on google analytics events, which isn't so easy to drill down into for individual events anyway, so not sure how useful it even is

src/app/shared/filter/filter.component.spec.ts
sandikbarr: yeah, if you select Last Month for example and do weekly, the 1st wasn't a Sunday in September, so it's making the date parameters match the dates that are gonna show up in the chart and table. I hadn't done anything with this at first and it felt weird to have the dates mismatch.

PR #65, opened 2017-10-27
use the CMS zoom parameter

sandikbarr:
to limit requests to only embedding the distributions


package.json
```diff
@@ -14,16 +14,16 @@
...
-    "@angular/common": "^4.0.0",
-    "@angular/compiler": "^4.0.0",
-    "@angular/core": "^4.0.0",
-    "@angular/forms": "^4.0.0",
-    "@angular/http": "^4.0.0",
-    "@angular/platform-browser": "^4.0.0",
-    "@angular/platform-browser-dynamic": "^4.0.0",
-    "@angular/router": "^4.0.0",
-    "@ngrx/store": "^4.0.3",
+    "@angular/animations": "^4.4.6",
```
sandikbarr: higher numbers are more better, right? I saw that adflow had a more recent angular version, so I added @latest here. Seemed safe enough since we're in active development.

src/app/app.component.ts
```diff
@@ -123,7 +123,7 @@ export class AppComponent implements OnInit, OnDestroy {
     podcast.doc.followItems('prx:stories', {
       per: podcast.doc.count('prx:stories'),
       filters: 'v4',
-      sorts: 'released_at: desc, published_at: desc'
```
sandikbarr: there's no reason to do this episode sort on the request since it's done in the reducer, one less burden for CMS

package.json
kookster: what could go wrong? 😸

Issue #57, opened 2017-10-24
number formatting


sandikbarr:
* Numbers should be right aligned.
* How should numbers be formatted in metrics? Should there be commas

PR #66, opened 2017-10-27
number formatting and alignment in table

sandikbarr:
#57 to right align numbers in table and format with thousands separator

I didn't do anything with the chart number formatting yet, ticket: https://github.com/PRX/styleguide.prx.org/issues/69

Also seeing this issue below with the table when scrolled horizontally. Ticket: https://github.com/PRX/metrics.prx.org/issues/67
screen shot 2017-10-27 at 5 45 34 pm

### DIFF
--- a/src/app/shared/pipes/large-number.pipe.ts
+++ b/src/app/shared/pipes/large-number.pipe.ts
@@ -1,0 +1,8 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({name: 'largeNumber'})
+export class LargeNumberPipe implements PipeTransform {
+  transform(value: number) {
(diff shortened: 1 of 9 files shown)


src/app/shared/pipes/large-number.pipe.ts
```diff
@@ -0,0 +1,8 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({name: 'largeNumber'})
+export class LargeNumberPipe implements PipeTransform {
+  transform(value: number) {
+    return Number(value).toLocaleString(undefined, {useGrouping: true});
```
gcampo88: nice! useGrouping is true by default, but it makes it quite clear what this is doing, so 👍

PR #96, opened 2017-11-28
don't include episodes scheduled for publishing that are not yet published

sandikbarr:
Episode that have been scheduled for publishing but are not yet published should not be included in metrics.


src/app/app.component.spec.ts
```diff
@@ -100,38 +139,23 @@ describe('AppComponent', () => {
...
+    dateInFuture.setDate(dateInFuture.getDate() + 1);
+    episodes = (<MockHalDoc>podcast.doc).mockItems('prx:stories', [
+      {id: 123, title: 'A Pet Talk Episode', publishedAt: new Date()},
+      {id: 123, title: 'A Pet Talk Episode', publishedAt: dateInFuture}
+    ]);
+    episodes.forEach(e => {
+      e.mockItems('prx:distributions',
+        [{kind: 'episode', url: 'https://feeder.prx.org/api/v1/episodes/42b4ad11-36bd-4f3a-9e92-0de8ad43a515'}]);
+    });
+    expect(comp.getEpisodePodcastDistribution).toHaveBeenCalledTimes(1);
```
gcampo88: hang on, are we dispatching these newly created episodes ever? i think the `comp.getEpisodePodcastDistribution` that was called once might be from the setup in the beforeEach, no?

sandikbarr: Ooo, that is a good catch. Let me look into it.

sandikbarr:
Sure enough, back to you @gcampo88!

Issue #83, opened 2017-11-27
Date & time formatting


farski:
Across the app I think we should make sure that we avoid any ambiguous date formats (eg 10/11/12) in favor of something more verbose when space permits (eg Oct. 11, 2012) or an unambiguous short form when constrained (2012-10-11 or similar)

I would also favor using AM/PM for times, regardless of where we land on time zone support, rather than 24 hour time.

sandikbarr:
What's left here is to update the styleguide to allow Input() of date format.

sandikbarr:
This should get resolved with the implementation of the new date range picker https://github.com/PRX/styleguide.prx.org/issues/78. Although it is not clear to me whether or not we've decided to drop commas from the proposed date formats. https://github.com/PRX/metrics.prx.org/issues/80#issuecomment-351190601

PR #112, opened 2017-12-04
show hourly data in local timezone

sandikbarr:
Formats the hourly data in the local timezone. Date formats have also been adjusted to match #83 (except for the datepicker, where we should make the date format an `Input()`.


src/app/downloads/downloads-chart.component.ts
```diff
@@ -122,11 +122,9 @@ export class DownloadsChartComponent implements OnDestroy {
         case INTERVAL_MONTHLY:
           return monthYearFormat;
         case INTERVAL_WEEKLY:
-          return dayMonthDateFormat;
         case INTERVAL_DAILY:
-          return dailyDateFormat;
+          return dayMonthDateFormat;
```
sandikbarr: The chart used to have the day of the week. This was based on user input that day of the week is useful info and that charts should start on either Sundays or Mondays rather than be rolling 7 days or whatever That same user contradicted himself in the last testing session and said that he wants rolling 7 day windows. Further, having the day of the week degrades somewhat for charts with more than 9 data points. It will only show the date on every other data point, so the start of the week is not marked, which I think is probably kinda what the user was getting at for wanting to know the day fo the week, so it didn't seem that useful. We might get some negative feedback on this set of changes because of removing the time pickers, so I figured it was as good a time as any to test out these proposed date and time formats.

src/app/downloads/downloads-table.component.ts
```diff
@@ -185,8 +187,7 @@ export class DownloadsTableComponent implements OnDestroy {
         case INTERVAL_DAILY:
           return dayMonthDateFormat(date);
         case INTERVAL_HOURLY:
-        case INTERVAL_15MIN:
-          return hourlyDateFormat(date);
+          return hourlyDateFormat(date).split(', ').join(',\n');
```
sandikbarr: This is what that `white-space: pre` formatting business is about. 😉

src/app/ngrx/actions/castle.action.creator.ts
```diff
@@ -1,11 +1,12 @@
-import { ActionTypes, ActionWithPayload } from './action.types';
+import { Action } from '@ngrx/store';
+import { ActionTypes } from './action.types';
 import { FilterModel, EpisodeModel, PodcastModel, MetricsType } from '../model';
 
 export interface CastleFilterPayload {
   filter: FilterModel;
 }
 
-export class CastleFilterAction implements ActionWithPayload<CastleFilterPayload> {
```
sandikbarr: `ActionWithPayload` came from the [migration guide](https://github.com/ngrx/platform/blob/master/MIGRATION.md#action-interface) and it did not pan out. It was causing TypeScript to complain about my ActionReducerMap, which was a red herring on that problem with the reducers not getting called last week.

src/app/ngrx/actions/index.ts
```diff
@@ -1,19 +1,18 @@
...
 
-export type All
-  = CmsPodcastsPayload
-  | CmsEpisodeGuidsPayload
-  | CastleFilterPayload
-  | CastlePodcastMetricsPayload
-  | CastleEpisodeMetricsPayload
-  | RouterNavigationPayload<RouterStateSnapshot>
-  | GoogleAnalyticsEventPayload;
+export type AllActions
```
sandikbarr: I _should_ be able to use this discriminating union in the reducers rather than having to use the individual Action classes. The problem is that TypeScript only seeing the properties on the CmsPodcastsAction when you give the reducer's action parameter the type `AllActions`. They are still recommending this though. I was using a later version of TypeScript at one point, but I downgraded to 2.4.0 because that's the earliest that Ngrx supports and Angular doesn't think it supports > 2.4.0.

src/app/shared/filter/date/custom-date-range.component.spec.ts
```diff
@@ -31,33 +31,13 @@ describe('CustomDateRangeComponent', () => {
...
-  });
-
   it('should not allow users to select dates more than 40 days apart when interval is hourly', () => {
     comp.interval = INTERVAL_HOURLY;
     comp.beginDate = beginningOfLastYearUTC().toDate();
     comp.endDate = endOfLastYearUTC().toDate();
     expect(comp.invalid).toContain('cannot be more than 40 days apart');
   });
-
-  it('should send google analytics event when begin or end time changes', () => {
```
cavis: Are we no longer sending this GA event?

src/app/ngrx/actions/action.types.ts
```diff
@@ -1,15 +1,8 @@
-import { Action } from '@ngrx/store';
-
 export const ActionTypes = {
   CMS_PODCASTS: 'CMS_PODCASTS',
-  CMS_EPISODE_GUID: 'CMS_EPISODE_GUID',
```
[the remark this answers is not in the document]
cavis: Or it was just un-used.

cavis:
Also ... object spread operator FTW!

Issue #117, opened 2017-12-12
episode chart selection


sandikbarr:
* [x] A paged table will be shown with the pages ordered by release date.
* [x] The page size is 10
* [x] Checkboxes will be included in the cells in the column titled "Episode." The podcast and each episode will have checkboxes. The checkboxes will toggle charting of each of the datasets. Checked means charted.
* [ ] The charted episodes and whether or not the podcast is charted will be saved in the url so that users can share the url to show the chart in the way that they've set it up and also so that we can create links into charts with single episodes charted.

Issue #109, opened 2017-11-29
Report, summary table, paging and episode selection


sandikbarr:
~@farski I'm also a little unclear what goes in the table on the different chart scenarios. On the one shown in the mockup, is that a listing of all of the podcasts episodes with the total downloads during the date range, release date, daily average, and all time downloads? Will that table change at all based on what type of chart is shown?~
@farski do averages change based on the interval selected? The mockup shows daily averages with the daily interval, so will those be hourly, weekly, or monthly with our other intervals?

A paged table will be shown with the pages ordered by release date. 
* [ ] The page size is 10
* [ ] On chart types that show individual episode data, there will be fancy checkboxes that also double as a legend to allow the user to toggle on or off each of the episodes from appearing in the chart
* [ ] A summary view with Total during the period, average, and all time downloads will be the default table view
* [ ] The user can select to expand out the table to also show the report breakdown containing each of the individual data points.
  * [ ] A Google Analytics event will be tied to this toggle so we can see how often users are looking at the reports

The table is not (yet?) sortable or searchable. However further questions we had for iterating on this approach is if there is some way to interact with the table, a filter or a go to page (more indicated by release date or episode title rather than requiring mental math with a page number), that allows the user to seek directly to an episode or set of episodes that they want to see that may have been released some time ago and further back in the pages.

sandikbarr:
implementation notes: to get the total all time downloads, Castle has to be running with rollups against BQ prod. `downloads.total` is available at the `/api/v1/episodes/:guid` or `/api/v1/podcasts/:id` endpoint.

PR #124, opened 2017-12-14
episode paging and selection

sandikbarr:
#117 This is the episode paging and selection part of epic #109


src/app/downloads/downloads.component.ts
```diff
@@ -38,11 +45,22 @@ export class DownloadsComponent implements OnInit, OnDestroy {
...
               public store: Store<any>,
-              private router: Router) {}
+              private router: Router,
+              private route: ActivatedRoute) {
+    route.params.subscribe(params => {
+      if (params['chartPodcast']) {
+        this.chartPodcast = params['chartPodcast'] === 'true';
+      }
+      if (params['episodes']) {
+        this.chartedEpisodes = params['episodes'].split(',').map(id => +id);
```
gcampo88: Seeing a thing where if you uncheck all episodes, and then recheck any of them, the last episode you unchecked will show as checked again. To solve I think you need an else this.chartedEpisodes = [] here.

src/app/ngrx/effects/cms.effects.ts
```diff
@@ -0,0 +1,95 @@
...
+          return series.followItems('prx:stories', {
+            page: payload.page,
+            per: EPISODE_PAGE_SIZE,
+            sorts: 'published_at: desc',
+            filters: 'v4',
+            zoom: 'prx:distributions'})
+          .flatMap((docs: HalDoc[]) => {
+            const chartedEpisodes = this.episodeMetrics.filter(e => e.charted).map(e => e.id);
+            // if none of the incoming episodes are already on the route, we want to add the first 5
+            const chartIncomingEpisodes = chartedEpisodes.filter(id => docs.map(doc => doc['id']).indexOf(id) !== -1).length === 0;
```
gcampo88: it felt little wonky to navigate around the pages and have the only the first 5 default-selected on every new page... but I don't have any better ideas!

src/app/ngrx/reducers/episode-metrics.reducer.ts
```diff
@@ -1,32 +1,65 @@
...
-      epIdx = state.findIndex(e => e.seriesId === seriesId && e.id === id);
-      if (epIdx > -1) {
-        episode = {...state[epIdx], guid};
-        setEpisodeMetrics(action.payload.filter.interval, action.payload.metricsType, episode, action.payload.metrics);
-        newState = [...state.slice(0, epIdx), episode, ...state.slice(epIdx + 1)];
-      } else {
-        episode = { seriesId, id, guid };
-        setEpisodeMetrics(action.payload.filter.interval, action.payload.metricsType, episode, action.payload.metrics);
-        newState = [episode, ...state];
+      if (action instanceof CastleEpisodeMetricsAction) {
```
gcampo88: I assume I'll learn from the Ultimate Angular videos why we're now checking instanceof? (Are we planning to use a single Action Type for multiple actions?) ... I should just watch the videos!

src/app/ngrx/reducers/index.ts
```diff
@@ -24,18 +26,23 @@ export const reducers: ActionReducerMap<RootState> = {
...
+
+export const selectPodcastState = createSelector(selectAppState, (state: RootState) => state.podcasts);
+export const selectPodcastEntities = createSelector(selectPodcastState, getPodcastEntities);
+export const selectPodcasts = createSelector(selectPodcastEntities, entities => {
+  return Object.keys(entities).map(seriesId => entities[parseInt(seriesId, 10)]);
+});
+export const selectPodcastsError = createSelector(selectPodcastState, getPodcastError);
+
+export const selectEpisodeState = createSelector(selectAppState, (state: RootState) => state.episodes);
+export const selectEpisodeEntities = createSelector(selectEpisodeState, getEpisodeEntities);
```
gcampo88: does this (and ditto for Podcast equiv) need to be exported if it's only used by another fn in here? Do you anticipate other places calling it?

src/app/ngrx/effects/cms.effects.ts
sandikbarr: Quality feedback. Noted. :+1: This is what was negotiated and landed on in the episode selection design decisions.

src/app/ngrx/reducers/index.ts
sandikbarr: I don't expect entities to be used outside of other selectors, but it's a pattern I'm working on following that these are exported. This entity pattern will allow me to turn the filter util functions into actual selectors; just the thing I was looking for initially to get smaller slices of state into selectors. With this entity pattern for example, we could have a selector that gives us a sorted list of episodes and remove that logic from components. I _think_ we should also be able to map data to the timeseries models that the chart uses in those selectors. It's a work in progress. and I'm still not quite sure how I'm going to fit the work in based on priority tbh, but that's the thought behind it.

src/app/downloads/downloads.component.ts
sandikbarr: Good catch. This stuff is really convoluted without doing router state composition.

src/app/ngrx/reducers/episode-metrics.reducer.ts
sandikbarr: I'm not sure you're gonna learn this from the videos. It's not initially straightforward or obvious to me at least, but the videos are setting up the action.payload to not have any additional levels of properties. If a new item is added to the state, say an episode, the episode's properties are on the action.payload rather than having an action.payload.episode at a child level. If it's an incoming response of a whole set of episodes, the payload is the array. They do this so that it can be assumed that payload is of type any. It seems like that eliminates the type checking benefits of having typed payloads. In order to follow the discriminating union pattern (our AllActions) we would need to be setting up our payloads similarly because of type safety. But if you want to have typed payloads, you can wrap these in instanceof checks to make TypeScript recognize the action and its payload type. I've had this open question about why the discriminating union pattern isn't working for a while; it's just that now I had to face it head on because I'm supporting more than one action type in some of the reducers. At least I had better examples to understand where I'd gotten off.

PR #127, opened 2018-01-02
Shared server lib

cavis:
Similar to PRX/publish.prx.org#492.

(Fixed upstream) ~~Still hitting a _slight_ issue where the dev server doesn't know ahead of time whether to include `scripts.bundle.js` or not.  (Determined by the `scripts: []` in `.angular-cli.json`).~~


env-example
```diff
@@ -17,5 +17,9 @@ NEW_RELIC_APP_NAME=
 NEW_RELIC_LICENSE_KEY=
 GA_KEY=
 
+# dev server
+PORT=4202
```
cavis: Note: you _MUST_ copy these to your actual `.env`, or things won't work.

The underlying `ng serve` doesn't know what actual HOST your browser is connecting to, as it's behind several proxies now!  And I think the default PORT=4200 will be used unless you have it in your ENV.

cavis:
OH THE COLORS!

env-example
```diff
@@ -11,11 +11,17 @@ CMS_TTL=
 AUTH_HOST=id.staging.prx.tech
 # for metrics.prx.dev
 AUTH_CLIENT_ID=ea0e04f7c85a8d4826dcf2747126a53a95014f51
```
sandikbarr: Is this a new auth id for https://metrics.prx.dev? It gives me a CORS error, but I can use the one I was previously using `nz8hAjerWIW5rudwAWPPd1zC1Lawt89HMUNVJpBa` and my callback url is https, no warnings or errors.

sandikbarr: yeah, `ea0e04f7c85a8d4826dcf2747126a53a95014f51` is the staging key. `nz8hAjerWIW5rudwAWPPd1zC1Lawt89HMUNVJpBa` is the production key.

Issue #118, opened 2017-12-12
chart toggle legend/fancy checkboxes


sandikbarr:
On chart types that show individual episode data, there will be fancy checkboxes that also double as a legend to allow the user to toggle on or off each of the episodes from appearing in the chart. When checked, the checkboxes will be colored the same as the dataset in the chart.

Issue #119, opened 2017-12-12
summary table/expandable report


sandikbarr:
* [x] A summary view with Total during the period, average, and all time downloads will be the default table view
* [x] The user can select to expand out the table to also show the report breakdown containing each of the individual data points.
* [x] A Google Analytics event will be tied to this expand button so we can see how often users are looking at the reports

gcampo88:
note to self: per slack discussion, the average should be the average per the selected interval. e.g. if the user has selected Daily, we should show Daily Average (as in mockup). 


design q: is this the design to work off of? @farski / @sandikbarr ? https://drive.google.com/file/d/18uV471p6ATZb0kuxmJj7YgR5ABZReL9L/view

sandikbarr:
@gcampo88 yup, that's the design but it's not showing the expandable report because I convinced @kookster and @farski we shouldn't take away the full report until there is a reports replacement, so that part of the design with the expandable button is mostly a vague idea in my head. With the ability to expand to the full report, I feel like the summary table doesn't need to take up the full width of the content area.

Implementation wise, I think we could put the summary stuff in the first "sticky" table and make the report the second scrolling table.

PR #131, opened 2018-01-04
Fancy checkboxes

cavis:
See #118


src/app/downloads/downloads-chart.component.ts
```diff
@@ -109,6 +107,8 @@ export class DownloadsChartComponent implements OnDestroy {
         // if we have episodes to combine with podcast total
         const episodeDatasets = this.episodeChartData.map(m => m.data);
         const allOtherEpisodesData: TimeseriesChartModel = {
+          // data: this.podcastChartData.data,
```
sandikbarr: can you remove this commented code?

src/app/downloads/downloads-table.component.css
```diff
@@ -23,17 +24,21 @@ td {
...
   overflow: hidden;
   text-align: left;
 }
 .scroll-x-wrapper {
-  max-width: calc(100% - 150px);
+  flex-grow: 1;
+  max-width: calc(100% - 200px);
   overflow-x: auto;
 }
+.scroll-x-wrapper > table {
```
sandikbarr: With this 100% wide, I dunno where we're gonna put the expandable report when this turns into the summary table with Gigi's ticket #119, but that's a design problem I'm just not getting very far with in discussion. I feel lucky I was able to convince them we shouldn't remove the report, shrug.

src/app/downloads/downloads-table.component.spec.ts
```diff
@@ -202,4 +206,13 @@ describe('DownloadsTableComponent', () => {
     fix.detectChanges();
     expect(de.query(By.css('em')).nativeElement.textContent).toContain('local timezone');
   });
+
+  it('toggles episode display when checkbox is clicked', () => {
```
sandikbarr: thanks for adding this test!

src/app/ngrx/effects/cms.effects.ts
```diff
@@ -39,6 +40,8 @@ export class CmsEffects {
                 seriesId: payload.podcast.seriesId,
                 title: doc['title'],
                 publishedAt: doc['publishedAt'] ? new Date(doc['publishedAt']) : null,
+                color: getColor(EPISODE_PAGE_SIZE, i),
+                // color: getShade(EPISODE_PAGE_SIZE, i),
```
sandikbarr: we'll never really be able to assign shading from here since we don't have the metrics, so let's remove this commented code

src/app/downloads/downloads-chart.component.ts
cavis: Whoops - will do.  I was experimenting with different colors/shades, and changing the graph type to "line", which required this other data.

src/app/downloads/downloads-table.component.css
cavis: That's to the left of both the chart and table, right?  Or maybe that's changed, per your comments.

Yeah, I think all of this table CSS will have to be readdressed in that ticket.  I thought about rewriting this whole table in flexbox ... but the possibility of failure was too much for me!

src/app/ngrx/effects/cms.effects.ts
cavis: Yeah, it won't work great unless the episodes are weighted by num-downloads.  I was thinking it might work _okay_ since more-recent episodes tend to have more downloads.  But will remove - this was just me experimenting with the colors anyways!

src/app/downloads/downloads-table.component.css
sandikbarr: Verbally, we discussed that the expandable report would expand to the right of the summary table. The summary table will take the place of this report and the report is the expandable bit. I think we are gonna want to redo the table in flexbox to make adaptive layout easier to deal with, but no need to do that today.

Issue #123, opened 2017-12-12
advanced range selector popup


sandikbarr:
The advanced range selector allows for the selection of an arbitrary date range. Any time the popup goes from being not displayed to displayed, it should reflect the current state of the app. Even in cases where the current range matches one of the presets, there is no active/highlighted state for the preset selectors in the popup. The beginning and ending date of the range are selected through calendar widgets, or through inputing dates in text fields. The popup also includes an interval selector.

The calendar widgets should always show the month of the currently selected day for each terminal of the range. The selected day is highlighted; the range is not. The user can scroll through months of each widget independently.

When changes are made to selections in the popup, they are not applied immediately. Instead, they are applied all at once when the user applies them, which dismisses the popup and invokes the changes.

Clicking outside the bounds of the popup dismisses the popup and trashes any changes.
![image](https://user-images.githubusercontent.com/2250413/33908964-a85f586e-df4f-11e7-8cce-b5dd0a7d8bb8.png)

PR #132, opened 2018-01-05
advanced range selector popup

sandikbarr:
#123 brings all the filtering together into an advanced popup


src/app/shared/menu/date/custom-date-range-dropdown.component.ts
```diff
@@ -1,48 +1,87 @@
...
-import { FilterModel, INTERVAL_HOURLY } from '../../../ngrx';
+import { FilterModel, INTERVAL_HOURLY, IntervalModel, IntervalList } from '../../../ngrx';
 import * as dateUtil from '../../util/date';
 import { GoogleAnalyticsEventAction } from '../../../ngrx/actions';
 
 @Component({
   selector: 'metrics-custom-date-range-dropdown',
   template: `
-    <div class="custom-date-range-dropdown" [class.open]="open">
+    <div class="dropdown" [class.open]="open">
```
sandikbarr: I havent made this dropdown structure a dropdown component with projected content because adding the custom styles for the dropdown-button and dropdown-content seems like it would just get unnecessarily complicated.

src/app/shared/menu/date/custom-date-range-dropdown.component.ts
```diff
@@ -1,48 +1,87 @@
...
+  intervalList = IntervalList;
+  userChoseRange: string;
+  CUSTOM_DATE = 'custom-date';
+  STANDARD_DATE = 'standard-date';
 
   constructor(public store: Store<any>) {}
 
   ngOnChanges() {
-    this.dateRange = this.filter;
+    this.tempFilter = this.filter;
```
cavis: Not sure I understand why you copy `this.filter -> this.tempFilter` both here and in the toggleOpen.  I guess I thought maybe just toggleOpen would be enough?

src/app/shared/menu/date/custom-date-range-dropdown.component.spec.ts
```diff
@@ -89,4 +91,25 @@ describe('CustomDateRangeDropdownComponent', () => {
     fix.detectChanges();
     expect(comp.invalid).toContain('dates in the past or present');
   });
+
+  it('keeps tempFilter standard range in sync with custom range', () => {
+    comp.onCustomRangeChange({from: dateUtil.beginningOfLastWeekUTC().toDate(), to: dateUtil.endOfLastWeekUTC().toDate()});
+    expect(comp.tempFilter.standardRange).toEqual(dateUtil.LAST_WEEK);
+  });
+
+  it('keeps tempFilter date range in sync with interval', () => {
```
cavis: Wooo tests!

src/app/shared/menu/date/custom-date-range-dropdown.component.ts
cavis: Like you'd have to use a bunch of deep selectors or something?

sandikbarr: No, like I'd want to style the outer components according to what I'm putting in there, not even deep selectors will solve that one

sandikbarr:
That blue emphasis font on the active "selected" standard range comes right from the styleguide. TBH, I've had a problem with that lack of contrast since we first started using it in Publish search paging. I consider that to be a much bigger question than this PR, so even though it bugs you, that's what our active/selected links look like.

sandikbarr:
Yes, we're keeping both the standard quick selector and the advanced popup. You'll have to talk @farski out of that one.

src/app/shared/menu/date/custom-date-range-dropdown.component.ts
sandikbarr: You're right, it didn't need this onChanges. Fixed.

cavis:
(waits patiently for build to pass)

sandikbarr:
aw, geez. it would help if I checked my tests...

PR #140, opened 2018-01-08
Move authorized CMS requests into effect

cavis:
Learn by ~failing~ doing.

- [x] Get the cms.auth requests out of AppComponent and into an Effect
- [x] Add a new account reducer for header
- [x] Make sure auth-token reloading (after login) re-triggers the account/podcasts actions
- [x] Effects testing


src/app/ngrx/effects/cms.effects.spec.ts
```diff
@@ -1,124 +1,208 @@
...
+
+    it('successfully loads the account', () => {
+      const accounts = auth.mockItems('prx:accounts', [
+        {name: 'TheAccountName', type: 'IndividualAccount', id: 111},
+        {name: 'DefaultName', type: 'DefaultAccount', id: 222}
+      ]);
+      const account = {id: 111, name: 'TheAccountName', doc: accounts[0]};
+      const action = new ACTIONS.CmsAccountAction();
+      const completion = new ACTIONS.CmsAccountSuccessAction({account});
+      actions$ = hot('-a', {a: action});
```
cavis: I have no idea if i'm doing these marbles correctly or not.

src/app/ngrx/effects/cms.effects.ts
```diff
@@ -1,63 +1,97 @@
...
 @Injectable()
 export class CmsEffects {
   episodeMetrics: EpisodeMetricsModel[] = [];
   routeParams: Params;
 
+  @Effect()
+  loadAccount$: Observable<Action> = this.actions$
+    .ofType(ACTIONS.ActionTypes.CMS_ACCOUNT)
+    .switchMap(() => {
+      return this.auth.token.first().mergeMap(token => {
```
cavis: Had to throw a `first()` in here to make sure this observable actually finishes, and doesn't re-run every time a new token is emitted.

src/app/ngrx/effects/cms.effects.ts
```diff
@@ -71,17 +105,17 @@ export class CmsEffects {
...
-            episode.guid = urlParts[urlParts.length - 1];
-          }
-        });
-    });
+  getEpisodeColor(episodeIndex: number): string {
+    return getColor(EPISODE_PAGE_SIZE, episodeIndex);
+  }
+
+  // if none of the incoming episodes are already on the route, we want to add the first 5
+  chartIncomingEpisodes(episodes: EpisodeModel[]) {
```
cavis: Ended up refactoring this somewhat, as it's called from a different point in the effect now.

src/app/ngrx/reducers/podcast.reducer.ts
```diff
@@ -41,6 +41,7 @@ export function PodcastReducer(state: PodcastState = initialState, action: AllAc
     case ActionTypes.CMS_PODCASTS: {
       return {
         ...state,
+        error: null,
```
cavis: Was having a problem where the "You are not logged in" message wasn't being cleared, when we retried loading the Podcasts.

src/app/ngrx/effects/cms.effects.spec.ts
sandikbarr: looks right to me, expect to get an observable of the success action with that account payload

src/app/ngrx/effects/cms.effects.ts
sandikbarr: 👍 that's not one I've had to use before

src/app/ngrx/effects/cms.effects.ts
```diff
@@ -94,4 +129,48 @@ export class CmsEffects {
     params['episodes'] = episodeIds.join(',');
     this.router.navigate([this.routeParams['seriesId'], 'downloads', this.routeParams['interval'], params]);
   }
+
+  private docToPodcast(doc: HalDoc): Observable<PodcastModel> {
```
sandikbarr: Looks like we can actually fully get rid of that `doc: HalDoc` property on the models now. I had originally put that in there because I was querying stories off the series doc, and I immediately hated it.

PR #151, opened 2018-01-10
Summary table

gcampo88:
#119 (and #157 incidentally)


src/app/downloads/downloads-table.component.ts
```diff
@@ -131,11 +81,25 @@ export class DownloadsTableComponent implements OnDestroy {
     this.podcastMetricsStoreSub = this.store.select(selectPodcastMetrics).subscribe((podcastMetrics: PodcastMetricsModel[]) => {
       this.podcastMetrics = findPodcastMetrics(this.filter, podcastMetrics);
       if (this.podcastMetrics) {
+        if (!this.podcastMetrics.allTimeDownloads) {
+          this.store.dispatch(new CastlePodcastAllTimeMetricsLoadAction({ filter: this.filter }));
```
gcampo88: Soliciting advice on this front: sometimes when the `CastlePodcastAllTimeMetricsLoadAction` is fired (from the ngOnInit and the constructor), `this.podcasts` is not yet defined in the Castle Effects. So I'm reissuing the fetch here if we get to this point but still don't have all time downloads.

src/app/ngrx/effects/castle.effects.ts
```diff
@@ -0,0 +1,64 @@
...
+export class CastleEffects {
+  podcasts: PodcastModel[] = [];
+
+  @Effect()
+  loadAllTimePodcastMetrics = this.actions$
+    .ofType(ActionTypes.CASTLE_PODCAST_ALL_TIME_METRICS_LOAD)
+    .map((action: CastlePodcastAllTimeMetricsLoadAction) => action.payload)
+    .switchMap((payload: CastlePodcastAllTimeMetricsLoadPayload) => {
+      const podcast = filterPodcasts(payload.filter, this.podcasts);
+      if (podcast) {
```
gcampo88: (see other comment -- sometimes this.podcasts is not yet populated, so this sends a "No podcasts yet" error and doesn't retry. Wasn't sure whether the best way to handle would be add retry logic or just dispatch the Load action again from the component -- went with the latter)

src/app/downloads/downloads-table.component.ts
```diff
@@ -131,11 +81,25 @@ export class DownloadsTableComponent implements OnDestroy {
...
+        }
         this.buildTableData();
       }
     });
   }
 
+  ngOnInit() {
+    if (this.filter) {
```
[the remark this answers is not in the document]
gcampo88: good call! 🌮 🌮  back atcha!

gcampo88:
@sandikbarr think it's ready for you again! good idea on the ngIf!

PR #155, opened 2018-01-11
date range summary

sandikbarr:


src/app/shared/menu/menu-bar.component.css
```diff
@@ -1,8 +1,17 @@
...
+  display: block;
   background-color: white;
 }
 
+.menu-bar {
+  display: flex;
+}
+
+.summary {
+  display: flex;
```
gcampo88: is this `flex` because of the later stuff we plan to add next to it?

sandikbarr: yup, yup

gcampo88: 👍

Issue #156, opened 2018-01-11
Calendar UI shifts when number of weeks change


farski:
![ezgif-4-c7efc9ea1a](https://user-images.githubusercontent.com/3483/34839529-e2747326-f6cf-11e7-8514-9a7f7f259540.gif)

pkarman:
@sandikbarr can you duplicate and estimate please?

sandikbarr:
This is a flex align issue that I thought I handled when I first implemented it. Oops, guess I didn't.

Issue #145, opened 2018-01-09
Odd dropdown animation


farski:
When the interval and preset quick select menus become visible they do they weird little expand animation, but (on Chrome, at least) just of the padding around the contents. The advanced selector menu only seems to do it after the firs time that it's become visible.

pkarman:
@sandikbarr can you duplicate this and give some guess as to estimate? I'm giving it a medium priority.

sandikbarr:
can do. I actually already found a typo in the transition and have this fixed in my current PR.

Issue #146, opened 2018-01-09
Scrunched hit zone for menubar items


farski:
I would expect the entire zone within the border to be clickable for these menu bar widgets

![image](https://user-images.githubusercontent.com/3483/34742745-386ed3a6-f555-11e7-9452-35733a4af181.png)

but the highlighted area around the edge is not active

![image](https://user-images.githubusercontent.com/3483/34742739-30b7ba06-f555-11e7-81fc-55c0d9373e29.png)

pkarman:
@sandikbarr can you duplicate and estimate please?

sandikbarr:
I'm grouping some changes to the dropdowns together and will have this fix included as well.

sandikbarr:
I lied. This problem is bigger than the dropdowns. Icon buttons have margin on them that padding can't seem to take the place of. I could make the margin clickable but that doesn't change the color of the icon on hover. I'm going to pull this back out of the sprint and into the backlog.

Issue #138, opened 2018-01-08
use RouterStateSerializer rather than FilterModel


sandikbarr:
* Process the routes with a RouterStateSerializer from Ngrx router-store
* Rather than FilterModel/CastleFilterAction and the more generic routeFromFilter, use specific actions to update the parameters such as RouteSeries, RouteEpisode(s), RouteDateRange, RouteStandardRange, RouteInterval, or when updating all params is possible with a single action, RouteAdvanced

PR #164, opened 2018-01-17
style series dropdown

sandikbarr:
Adds the dropdown style to the podcasts navigation. I also tried to address some other dropdown issues so we could group some of these smaller ones for less context switching and have made comments about those specific changes.

One thing I tried to address but backed out of was clicking other dropdowns while one is open only clicks the overlay, so if you were trying to open another dropdown, you've gotta double click. I had a fix for this worked up to pass the click through the dropdown, but if that click happened to land on the current dropdown's dropdown button, the dropdown closed itself then re-opened. It felt super hacky and like I was really digging a hole for myself, so I yanked it. (Kind of off topic, but I also looked into using Components Mixins while looking at this problem, and people who are more experienced than me think it's a bad idea even if you can get it to work.) It bothers me that the dropdown component seems like it could be more DRY. I thought about using content project (like the hero component for example) but I also need to be able style each of the dropdown buttons separately and override some of that CSS for individual dropdowns (the opposite of deep selectors, I would need to style the elements that I'm projecting into.) I haven't quite found the right solution there.


src/app/shared/menu/date/custom-date-range-dropdown.component.css
```diff
@@ -24,7 +24,7 @@
 
 .ranges {
   display: flex;
-  align-items: center;
+  align-items: flex-start;
```
sandikbarr: fixes https://github.com/PRX/metrics.prx.org/issues/156

src/app/shared/menu/dropdown.css
```diff
@@ -43,8 +44,16 @@
...
   padding: 20px;
   max-height: initial;
-  transition: max-height;
-  transition: 0.5s;
+  overflow-y: auto;
+}
+.open .dropdown-content.rollout {
+  height: auto;
+  max-height: 500px;
+  transition: max-height 0.2s;
```
sandikbarr: fixes https://github.com/PRX/metrics.prx.org/issues/145 (those were meant to be `transition-property` and `trasition-duration` oops, and 0.5s was way too slow)

src/app/shared/menu/date/custom-date-range-dropdown.component.css
cavis: :+yep:

src/app/shared/podcast-nav/podcast-nav-dropdown.component.ts
```diff
@@ -0,0 +1,40 @@
...
+
+@Component({
+  selector: 'metrics-podcast-dropdown',
+  template: `
+    <div *ngIf="podcasts?.length > 1" class="dropdown" [class.open]="open">
+      <div class="overlay" (click)="toggleOpen()"></div>
+      <div class="dropdown-button">
+        <button (click)="toggleOpen()" >{{ selectedPodcast?.title }}<span class="down-arrow"></span></button>
+      </div>
+      <div class="dropdown-content rollout">
```
cavis: Can we put some sort of max-height on this?  Noticing it gets pretty long when I'm logged in as admin.  (Granted, this isn't the normal use-case ... but if the dropdown-height > page-height, you can't get to the bottom items.)

src/app/shared/podcast-nav/podcast-nav-list.component.ts
```diff
@@ -0,0 +1,26 @@
+import { Component, Input, Output, EventEmitter } from '@angular/core';
+import { PodcastModel } from '../../ngrx';
+
+@Component({
+  selector: 'metrics-podcasts-nav-list',
```
cavis: Plural "podcasts" vs singular "podcast" in the component name?

src/app/shared/podcast-nav/podcast-nav.component.ts
```diff
@@ -0,0 +1,45 @@
...
+  selectedInterval: IntervalModel;
+  selectedChartType: ChartType;
+  selectedPodcast$: Observable<PodcastModel>;
+  podcasts$: Observable<PodcastModel[]>;
+
+  constructor(private store: Store<any>,
+              private router: Router) {}
+
+  ngOnInit() {
+    this.filterSub = this.store.select(selectFilter).subscribe((filter: FilterModel) => {
```
cavis: Part of me wishes there weren't any subscribes at all in here (just like ngrx promised!).  But another part of me doesn't know how to accomplish that in this case.

cavis: Maybe make some `selectFilterInterval` and `selectFilterChartType` selectors, and then passing those down a level?  Hrmm... but I guess you'd have to move the routing down a level, which isn't ideal either.

src/app/shared/menu/dropdown.css
cavis: I'm in no hurry!

src/app/shared/podcast-nav/podcast-nav-dropdown.component.css
```diff
@@ -0,0 +1,7 @@
+.dropdown-button, .open .dropdown-button {
```
cavis: `<ui-opinion>`I do sort of wish there was more left/right padding on this button.  But that could just be me`</ui-opinion>`

src/app/shared/podcast-nav/podcast-nav-dropdown.component.ts
sandikbarr: These rollout dropdowns with transition have a max height of 500px. https://github.com/PRX/metrics.prx.org/pull/164/files#diff-54a5c8e229b3bb725603f6b52bcff45dR51 I could make that shorter...? I took the transition off the popup one and set max-height to initial because I didn't want to constrain it if I didn't have to.

cavis: Ah, okay, missed that (was testing with window < 500px tall).  As long as it has one, I'm good!

src/app/shared/podcast-nav/podcast-nav-list.component.ts
sandikbarr: oops! missed a spot in the rename to podcast-no-s-mav, will update!

src/app/shared/podcast-nav/podcast-nav-dropdown.component.css
sandikbarr: well, since I can while I'm hesitating with whether I like the shared+override CSS approach... what the heck, I can and I will. I agree it looks better with padding up there. We can figure out the menu bar padding details when we work on mobile layout.

src/app/shared/podcast-nav/podcast-nav.component.ts
sandikbarr: I do actually have a plan for getting rid of these. I think routing should be actions. Right now for example with the menu, I'm passing any routing up to the downloads component with EventEmitters, and I think those could be handled as actions by an effect rather than handled in a container component. https://github.com/PRX/metrics.prx.org/issues/138 Also I think filter being a RouteStateSerializer would be much cleaner.

src/app/shared/podcast-nav/podcast-nav-dropdown.component.css
sandikbarr: ugh this button padding/margin problem is a little dross, might not fix that icon button https://github.com/PRX/metrics.prx.org/issues/146 but I'm gonna keep working on it and will update in this PR.

sandikbarr: This icon button is a whole other issue, but I adjusted the other buttons to use padding instead of margin (I built the icon dropdown first and was reducing the styleguide btn-icon margin, still using margin with these other buttons was an oversight on my part,) and I added additional padding to the podcast-nav.

src/app/shared/podcast-nav/podcast-nav.component.ts
cavis: Yeah, that would make sense.  Good sense!

src/app/shared/podcast-nav/podcast-nav-dropdown.component.css
cavis: 👍

Issue #174, opened 2018-01-19
Episode title are clipped in table with single line chart


farski:
In most other cases long titles get truncated with `…`

![image](https://user-images.githubusercontent.com/3483/35162090-298737ee-fd10-11e7-959b-11ed8639282d.png)

Issue #74, opened 2017-11-09
spinner doesnt stop for series with no published episodes


sandikbarr:
On application load, if the selected series has no published episodes, the spinner does not stop spinning.

Issue #141, opened 2018-01-08
getChartData selector


sandikbarr:
A chart data selector would use the router state against the metrics stores to build the chart data including the podcast and/or episode data when charted as well as order the episode data according to total when the chart type is stacked.

pkarman:
can we bump this out since not user-facing?

sandikbarr:
This is a larger refactor that will follow the RouterStateSerializer but it's not necessary that it happen directly after the RouterStateSerializer, so yes, this one makes sense to bump. It's gonna simplify the chart component a lot and will make maintenance and enhancements so much easier to handle, so it is high value and we should keep at least medium priority.

PR #189, opened 2018-01-30
Fix loading issues

sandikbarr:


src/app/downloads/downloads-table.component.css
```diff
@@ -58,14 +58,19 @@ table.scroll-x th:first-child, table.scroll-x td:first-child {
...
-prx-checkbox, #episodeTitle button, #podcastTitle button {
+prx-checkbox {
   position: relative;
   margin-top: -4px;
   top: 4px;
   width: 100%;
   overflow: hidden;
   text-overflow: ellipsis;
 }
+.title {
```
sandikbarr: Fixes #174. The title text still isn't quite aligned with the other table columns, but let's worry about that when we can replace `<table>`

src/app/downloads/downloads.component.ts
```diff
@@ -2,26 +2,26 @@ import { Component, OnInit, OnDestroy } from '@angular/core';
...
 
 @Component({
   template: `
-    <prx-spinner *ngIf="isPodcastLoading || isEpisodeLoading" overlay="true"
+    <prx-spinner *ngIf="loading$ | async" overlay="true"
                  loadingMessage="Please wait..."></prx-spinner>
     <router-outlet name="sidenav"></router-outlet>
-    <section class="content" *ngIf="podcasts">
-      <metrics-menu-bar *ngIf="!isPodcastLoading && !isEpisodeLoading"></metrics-menu-bar>
+    <section class="content" *ngIf="loaded$ | async">
```
sandikbarr: Uses loading$ and loaded$ fixes #74. (I was testing on a series without a feederId this morning, and that was causing other swallowed errors that I was not realizing.)

src/app/ngrx/effects/cms.effects.ts
```diff
@@ -85,12 +92,16 @@ export class CmsEffects {
...
-          });
+        if (docs.length) {
+          return Observable.forkJoin(docs.map((doc, index) => this.docToEpisode(doc, seriesId, index, pageNum)))
+            .map(episodes => episodes.filter(e => e && e.guid))
+            .map(episodes => {
+              this.chartIncomingEpisodes(episodes);
+              return new ACTIONS.CmsPodcastEpisodePageSuccessAction({episodes});
+            });
+        } else {
+          return Observable.of(new ACTIONS.CmsPodcastEpisodePageSuccessAction({episodes: []}));
```
sandikbarr: This success action when empty plus loading$ and loaded$ instead of downloads.component tracking loading is fixing #74

src/app/ngrx/reducers/index.ts
```diff
@@ -6,8 +6,8 @@ import { EpisodeReducer } from './episode.reducer';
 import { PodcastMetricsReducer, PodcastMetricsModel } from './podcast-metrics.reducer';
 import { EpisodeMetricsReducer, EpisodeMetricsModel } from './episode-metrics.reducer';
 import { AccountState, getAccountEntity, getAccountError } from './account.reducer';
-import { PodcastState, getPodcastEntities, getPodcastError } from './podcast.reducer';
-import { EpisodeState, getEpisodeEntities } from './episode.reducer';
+import { PodcastState, getPodcastEntities, getPodcastsLoaded, getPodcastsLoading, getPodcastsError } from './podcast.reducer';
+import { EpisodeModel, EpisodeState, getEpisodeEntities, getEpisodesLoaded, getEpisodesLoading, getEpisodesError } from './episode.reducer';
 import { CustomRouterReducer } from './router.reducer';
```
sandikbarr: selectors should have some kind of logical separation and grouping and not all just live in this index file, maybe when we can get more refactoring done, but right now this index file isn't holding anything up

src/app/ngrx/reducers/podcast-metrics.reducer.ts
```diff
@@ -54,8 +84,8 @@ export function PodcastMetricsReducer(state: PodcastMetricsModel[] = initialStat
         return newState;
       }
       break;
-    case ActionTypes.CASTLE_PODCAST_ALL_TIME_METRICS_SUCCESS:
-      if (action instanceof CastlePodcastAllTimeMetricsSuccessAction) {
+    case ACTIONS.ActionTypes.CASTLE_PODCAST_ALL_TIME_METRICS_SUCCESS:
```
sandikbarr: I considered doing a loading/loaded status for all time metrics, but we don't have the podcast seriesId for the load action (chart component that calls the load all time actions is complicated and needs some selector love #141). all time metrics don't cause the same jumpiness in the UI, so I'm ok with letting these fill in as they load.

src/app/downloads/downloads.component.ts
```diff
@@ -2,26 +2,26 @@ import { Component, OnInit, OnDestroy } from '@angular/core';
...
-import { selectRouter, selectEpisodes, selectPodcasts } from '../ngrx/reducers';
+import { selectRouter, selectEpisodes, selectPodcasts, selectLoading, selectLoaded, selectErrors } from '../ngrx/reducers';
 import { filterPodcastEpisodePage } from '../shared/util/metrics.util';
 import * as dateUtil from '../shared/util/date';
 import { isPodcastChanged, isBeginDateChanged, isEndDateChanged, isIntervalChanged } from '../shared/util/filter.util';
 
 @Component({
   template: `
-    <prx-spinner *ngIf="isPodcastLoading || isEpisodeLoading" overlay="true"
+    <prx-spinner *ngIf="loading$ | async" overlay="true"
```
sandikbarr: I don't have any hard metrics on this at all, but seems like this was also a good move for improving browser performance. My fan isn't so loud anymore when the page is loading. There was a lot of data being filtered and worked through and page draws and charts generated every time a new piece of data came in.

src/app/ngrx/effects/routing.effects.ts
```diff
@@ -196,11 +203,25 @@ export class RoutingEffects {
...
     }
 
-    this.router.navigate([
-      combinedRouterState.podcastSeriesId,
-      combinedRouterState.metricsType,
-      combinedRouterState.chartType,
-      combinedRouterState.interval.key,
-      params]);
+    switch (combinedRouterState.metricsType) {
+      case METRICSTYPE_DOWNLOADS:
```
cavis: Is this to keep the chartType and interval out of the URL when they're the default?  Not entirely sure I follow.

src/app/ngrx/reducers/episode-metrics.reducer.ts
```diff
@@ -21,28 +22,57 @@ const episodeIndex = (state: EpisodeMetricsModel[], id: number, seriesId: number
...
-      if (action instanceof CastleEpisodeChartToggleAction) {
+    case ACTIONS.ActionTypes.CASTLE_EPISODE_METRICS_SUCCESS:
+      if (action instanceof ACTIONS.CastleEpisodeMetricsSuccessAction) {
+        const { id, seriesId, guid, page, metricsPropertyName, metrics } = action.payload;
+        const epIdx = episodeIndex(state, id, seriesId);
+        let episode: EpisodeMetricsModel, newState: EpisodeMetricsModel[];
+        if (epIdx > -1) {
+          episode = {...state[epIdx], id, seriesId, guid, page, loading: false, loaded: true};
+          episode[metricsPropertyName] = metrics;
+          newState = [...state.slice(0, epIdx), episode, ...state.slice(epIdx + 1)];
```
cavis: Is there a reason to store the episodes in an array instead of an object (keyed by guid)?

src/app/ngrx/reducers/index.ts
```diff
@@ -88,5 +96,51 @@ export const selectPodcastMetricsFilteredTotal = createSelector(selectPodcastMet
...
+const errorType = (code) => code === 401 ? 'Authorization' : 'Unknown';
+export const selectPodcastMetricsError = createSelector(selectPodcastMetrics, (metrics: PodcastMetricsModel[]) => {
+  return metrics.filter(m => m.error).map(m => {
+    return `${errorType(m.error.status)} error occurred while requesting podcast metrics`;
+  });
+});
 
 export const selectEpisodeMetrics = createFeatureSelector<EpisodeMetricsModel[]>('episodeMetrics');
+export const selectEpisodeMetricsLoading = createSelector(selectEpisodeMetrics, (metrics: EpisodeMetricsModel[]) => {
+  return metrics.some((m: EpisodeMetricsModel) => m.loading);
```
cavis: Ah, I get it.  So the `CastleEpisodeMetricsLoadAction`s are dispatched individually.  But the loading state is across all of those.

src/app/ngrx/reducers/index.ts
```diff
@@ -88,5 +96,51 @@ export const selectPodcastMetricsFilteredTotal = createSelector(selectPodcastMet
...
+  return metrics.every((m: EpisodeMetricsModel) => m.loaded || m.loaded === undefined);
+});
+export const selectEpisodeMetricsError = createSelector(selectEpisodeMetrics, (metrics: EpisodeMetricsModel[]) => {
+  return metrics.filter(m => m.error).map(m => {
+    return `${errorType(m.error.status)} error occurred while requesting episode metrics for ${m.guid}`;
+  });
+});
+
+export const selectCmsLoading = createSelector(selectEpisodesLoading, selectPodcastsLoading, (episodes, podcasts) => episodes || podcasts);
+export const selectCastleLoading = createSelector(selectEpisodeMetricsLoading, selectPodcastMetricsLoading,
```
cavis: Ahhh... yes.  And this aggregates podcast + all-episodes loading state.  All the things!

src/app/ngrx/reducers/index.ts
cavis: Yeah, agreed.  This file is getting a bit out of control ... but not worth tackling amidst all this other stuff.

src/app/ngrx/reducers/podcast-metrics.reducer.ts
cavis: agreed

src/app/ngrx/actions/castle.action.creator.ts
```diff
@@ -1,29 +1,91 @@
...
+  error: any;
+}
+
+export class CastlePodcastMetricsFailureAction implements Action {
+  readonly type = ActionTypes.CASTLE_PODCAST_METRICS_FAILURE;
+
+  constructor(public payload: CastlePodcastMetricsFailurePayload) {}
+}
+
+export interface CastleEpisodeMetricsLoadPayload {
```
sandikbarr: I should also explain why I moved away from using the models on these payloads. That's the TypeScript issue that requires we wrap our switch cases in those redundant if instance of checks. I realized that's what Todd Motto has been doing to get around it. The generic Action type has a payload, so you can destructure the payload in the reducer to get at the properties. Any deeper than one level past the payload, and you've got to use all those dumb instance of checks.

src/app/ngrx/effects/castle.effects.ts
```diff
@@ -2,51 +2,106 @@ import { Injectable } from '@angular/core';
...
+      return this.castle.followList('prx:episode-downloads', {
+        guid,
+        from: beginDate.toISOString(),
+        to: endDate.toISOString(),
+        interval: interval.value
+      }).map(
+        metrics => {
+          return new ACTIONS.CastleEpisodeMetricsSuccessAction({
+            seriesId,
+            page,
```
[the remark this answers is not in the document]
cavis: Hrmm... yeah, I'm not sure I answered this for myself.  Is this just to make the pager work?  A single episode wouldn't be on more than 1 page, right?

src/app/downloads/downloads.component.ts
```diff
@@ -210,65 +197,30 @@ export class DownloadsComponent implements OnInit, OnDestroy {
...
-        metricsPropertyName: getMetricsProperty(this.routerState.interval, this.routerState.metricsType),
-        metrics: metrics[0]['downloads']
-      }));
-      this.googleAnalyticsEvent('load', metrics[0]['downloads'].length);
-    }
-  }
-
-  googleAnalyticsEvent(gaAction, value) {
-    this.store.dispatch(new GoogleAnalyticsEventAction({gaAction, value}));
+    this.store.dispatch(new ACTIONS.CastlePodcastMetricsLoadAction({
```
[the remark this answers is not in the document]
sandikbarr: Yeah, I tried to make the changes to where the components didn't have to dispatch loading, but we're just not there yet. I've been asked to postpone non-blocking refactoring until after the beta release, so we can revisit.

src/app/ngrx/effects/castle.effects.ts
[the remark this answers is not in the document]
sandikbarr: Page is not really used right now. I think that my thought on including the page here was eventually a chart data selector could get both the episodes and the metrics by the page number? I seem really confident about that don't I?

src/app/ngrx/effects/routing.effects.ts
sandikbarr: yes, exactly that. On prod right now, a page refresh on demographics or traffic sources will through a no routes match error because it includes the chart type and interval, segments that don't exist in their routes. Doh.

src/app/ngrx/reducers/episode-metrics.reducer.ts
sandikbarr: Fewer and smaller code changes. I wouldn't have the go ahead to change any of this stuff if the loading issues didn't make the release.

src/app/ngrx/reducers/index.ts
sandikbarr: Yes, so the spinner spins on loading$ until all individual episode metrics are not loading (and episodes and series and podcast metrics.)

src/app/ngrx/reducers/index.ts
sandikbarr: yep, needed to compose it down to a single observable for `| async` on the spinner

sandikbarr:
> "couldn't podcast/episode state be an object rather than array?"

I think yes, but maybe not in the same way as the cms podcast and episode reducers. The reason the loading/loaded state can be at the top level for cms reducers is because only one request is made at a time, so you'll get load and then either a success or failure. But with metrics, the episode metrics are requested individually one after another to fill out the page, so I had to put the loading/loaded state with each individual entry rather than at the top. Hmm, I just did the same with podcast metrics for consistency I guess but could have loading/loaded/error at the top there because we're only ever looking at one at a time.

I do think that having a top level error on metrics might be one case for turning the state into an object. I'm not capturing that "No selected podcast yet" failure from the loadAllTimePodcastMetrics$ effect anywhere for example. That does seem more like a "this component setup is complicated and error prone" issue rather than one we'd be living with long term. I'm really hoping to make more moves on reducing complexity in components after the beta release, but this is where we are.

I also think that the object/entity pattern is nicer to work with in selectors. In a past life, when I was using Redux, I would have made another level of reducer and used combineReducers with the entity pattern so that the episodes, podcasts, and metrics reducers would be working on a single entry rather than the whole set. I haven't quite translated that into ngrx patterns though, so I'm gonna keep reading up on it.

Issue #106, opened 2017-11-29
Podcast Profile section


sandikbarr:
A podcast profile section is shown in the upper right. This ticket includes the basic info. Performance info will be added with #144
 * [x] A rounded (border radius 50%, similar to the navuser image) series image
 * [x] The Series title on a single line with overflow ellipsis
 * [x] The Most Recent Episode title on a single line with overflow ellipsis
 * [x] Most recent episode title is link to multi-line chart with only the single episode charted
![screen shot 2017-11-29 at 1 56 39 pm](https://user-images.githubusercontent.com/2250413/33396113-315778ec-d50d-11e7-830c-ffd35413ddf6.png)

sandikbarr:
why is the podcast profile image a circle when podcast images are square by design? (at least I assumed that's what the blue dot is)

pkarman:
Is this really high? I think we should bump this and #144 unless there's a functional gap that users will notice.

PR #193, opened 2018-02-01
Podcast profile

cavis:
See #106.


src/app/ngrx/reducers/index.ts
```diff
@@ -76,6 +76,18 @@ export const selectEpisodesError = createSelector(selectEpisodeState, getEpisode
 export const selectSelectedPageEpisodes = createSelector(selectEpisodes, selectPageRoute, (episodes: EpisodeModel[], page: number) => {
   return episodes.filter(episode => episode.page === page);
 });
+export const selectMostRecentEpisode = createSelector(selectSelectedPodcast, selectEpisodes, (pod: PodcastModel, eps: EpisodeModel[]) => {
+  // TODO: if you didn't start on the first page, this isn't the most recent
```
cavis: Any ideas on this one?  If the route starts on `page=2`, you'll never know about the most recent 10 episodes.  Bummer!

Maybe the next PR on this (to pull in the all-time-metrics) should just have an independent CMS call to get the 1-most-recent episode?

src/app/ngrx/reducers/index.ts
```diff
@@ -76,6 +76,18 @@ export const selectEpisodesError = createSelector(selectEpisodeState, getEpisode
...
+  let recent: EpisodeModel;
+  eps.forEach(episode => {
+    if (episode.seriesId === pod.seriesId) {
```
cavis: I'm not entirely clear on why i have to do this.  The above `selectSelectedPageEpisodes` uses the same `selectEpisodes` ... but it doesn't have to filter by seriesId.  What the what?

src/app/shared/profile/profile.component.ts
```diff
@@ -1,8 +1,64 @@
...
 @Component ({
   selector: 'metrics-profile',
-  template: ``
+  styleUrls: ['profile.component.css'],
+  template: `
+    <div class="profile">
+
+      <div class="info">
+        <div class="logo">
+          <prx-image [imageDoc]="podcast?.doc"></prx-image>
```
sandikbarr: Hmm, the image isn't updating for me when I select a different series

src/app/ngrx/reducers/recent-episode.reducer.ts
```diff
@@ -0,0 +1,68 @@
...
+  ActionTypes,
+  CmsRecentEpisodeSuccessAction,
+  CmsRecentEpisodeFailureAction,
+  AllActions
+} from '../actions';
+import { EpisodeModel } from './episode.reducer';
+import { HalDoc } from 'ngx-prx-styleguide';
+
+interface RecentEpisodeLookup {
+  [seriesId: number]: EpisodeModel;
```
sandikbarr: Recent episodes keyed by series id makes very good sense! 🌮

src/app/shared/nav/nav-menu.component.ts
```diff
@@ -62,4 +70,12 @@ export class NavMenuComponent {
       }
     });
   }
+
```
sandikbarr: Really just thinking out loud here. This is excellent work and does exactly what it is intended to. I do find myself wondering if it would be helpful to change at least the begin date to the release date of this episode, maybe even do a switch to hourly with a shorter date range. It's a good idea as a navigation tool to have an easy link to the latest episode performance since it seems that is what users are primarily interested in, just feel like it's falling a little short on the intended goal when most of the chart is often at zero from the default date range. I've often thought the default date range could be based on the release schedule, which might be a bigger idea that we are ready to tackle with the UX.

sandikbarr:
update from the styleguide should fix that image loading issue then ready to merge!

Issue #226, opened 2018-02-05
URL filling up with selected episodes


cavis:
Feature + bug.  As you navigate around (using the paging / podcast-dropdown), more and more episode ids end up in the URL.  Eventually I've got 100+ of `;episodes=1234,5678`.  Keeps track of your which-episode-selected-state ... but you may end up sharing a gigantic url that reflects more than the state of what's currently rendered.

I would expect the `;episodes=` to reset every time you (1) change podcasts or (2) change pages.

gcampo88:
that would help w/ this too https://github.com/PRX/metrics.prx.org/issues/202

sandikbarr:
The intention is to save the state of episode toggling from other pages you have visited. I think I'm the only one who had any strong opinions about that, sounds like the majority wants to clear it out for anything not in view.

pkarman:
+1 to clearing state on page or podcast change.

Issue #206, opened 2018-02-05
Strange episode selection behavior


farski:
Steps to reproduce:

1. Load Showcase metrics
1. Select _Secrets #4_ from Most Recent in profile
1. Turn on _Secrets #1_ in table
1. Turn off _#4_ in table
1. Turn off _#1_ in table
1. Turn on #4 in table

Only one thing should be selected at this point, but _#1_ and _#4_ will be

pkarman:
I can reproduce.

farski:
Also seems to happen any time it goes from 0-selected-episodes to 1 selected episode, and the selected episode isn't the episode that was selected prior to no episodes being selected

(🤔  I think that made sense)

Issue #204, opened 2018-02-05
Navigating to page 1 should reset to default episode selection


gcampo88:
If you have ever clicked on the "most recent" link in the podcast profile, then when you are on page 1 of the table, only that one episode will be charted. It's strange to navigate to other table pages and have the default 5 shown, then come back to page 1 and have only the one episode shown. Seems like it'd be better to restore the default 5 when navigating to page 1.

pkarman:
Related to #206 ?

Issue #202, opened 2018-02-05
Should clear out page info when navigating to different podcast?


gcampo88:
Currently if you navigate from one show to another, paging + timeframe + interval info is retained as you go to the new podcast. Feels like it would be better at least to wipe out the page info, so your view of the new show isn't page 9 or whatever on first look. 

** especially bad if there is no page X on the podcast you navigate to --  you get no chart and a one-line table, looks like a mistake.

pkarman:
+1 can reproduce.

Managing state is a Really Hard problem. Seems like a few tickets opened around this UX. I expect we'll need to iterate a few times.

PR #232, opened 2018-02-06
resets episode state on podcast or page change

sandikbarr:
#226, #202, #204, and #206


src/app/ngrx/reducers/episode-metrics.reducer.ts
```diff
@@ -122,25 +107,26 @@ export function EpisodeMetricsReducer(state: EpisodeMetricsModel[] = initialStat
...
-          if (epIdx > -1) {
-            episode = {id, seriesId: podcastSeriesId, ...newState[epIdx], charted: true};
-            newState = [...newState.slice(0, epIdx), episode, ...newState.slice(epIdx + 1)];
-          } else {
-            episode = {id, seriesId: podcastSeriesId, charted: true};
-            newState = [episode, ...newState];
-          }
-        });
-        return newState;
+      // update existing entries
```
sandikbarr: at least we're now a step closer to moving this state out of here and only keeping it on the route

gcampo88:
fixes all the things! 🌮 🌮 . can now delete `CASTLE_EPISODE_CHART_TOGGLE` & `CASTLE_PODCAST_CHART_TOGGLE` from action.types.ts as well! want to do that before merge?

sandikbarr:
Good catch on removing those consts. Will wait for build confirm then merge.

Issue #214, opened 2018-02-05
Truncated episode titles could have full title in their title attribute


farski:
eg 

`<button title="The Polybius Conspiracy #7 – The Final Level" _ngcontent-c13="" class="btn-link title">The Polybius Conspiracy #7 – The Final Level</button>`

to provide helpful tooltips

Issue #201, opened 2018-02-05
Reduce the size of disclosure triangles


farski:
Currently these are using an 8px border

![image](https://user-images.githubusercontent.com/3483/35816613-21ae8f20-0a69-11e8-8716-ce410befb4f3.png)

They could come down to 6px

![image](https://user-images.githubusercontent.com/3483/35816711-701a3d8a-0a69-11e8-90ef-333a3c113035.png)

Issue #190, opened 2018-01-31
average math is wrong


sandikbarr:
Averages for partial weeks and months should be weighted to how many days in the week or month are included.

Issue #221, opened 2018-02-05
Lag in All Time Downloads 


palomao:
There's a lag the further back you go in a series for All Time Downloads to load.
![screen shot 2018-02-05 at 12 00 10 pm](https://user-images.githubusercontent.com/17989422/35817806-5ec2dfc6-0a6c-11e8-9354-1571c50c85dc.png)

sandikbarr:
I've added loading status to the all time/performance metrics, so we'd just need to add a check for it in the loading reducer in order to display the spinner until the all time downloads loading has completed. Got the groundwork laid in Bernard when adding performance metrics, so now we just have a small change left to implement.

PR #240, opened 2018-02-13
profile performance info

sandikbarr:
adds the profile performance info.
also included:
  * remove averages #190
  * reduce size of disclosure triangles #201
 * truncated episode titles have episode title as title attribute #214


src/app/downloads/downloads-table.component.html
```diff
@@ -34,22 +33,20 @@
...
-          <td>{{podcastTableData.avgPerIntervalForPeriod | largeNumber}}</td>
           <td *ngIf="podcastTableData.allTimeDownloads !== undefined">{{podcastTableData.allTimeDownloads | largeNumber}}</td>
           <td *ngIf="podcastTableData.allTimeDownloads === undefined"></td>
         </tr>
         <tr *ngFor="let episode of episodeTableData">
           <td>
             <prx-checkbox *ngIf="showEpisodeToggles; else episodeTitle"
               small [checked]="episode.charted" [color]="episode.color"
-              (change)="toggleChartEpisode(episode, $event)">{{episode.title}}</prx-checkbox>
+                          (change)="toggleChartEpisode(episode, $event)"><span [title]="episode.title">{{episode.title}}</span></prx-checkbox>
```
sandikbarr: #214 it was said that it's ok to include the title even when the title is not truncated

src/app/downloads/downloads-table.component.html
```diff
@@ -34,22 +33,20 @@
...
             <ng-template #episodeTitle>
-              <button class="btn-link title" (click)="onChartSingleEpisode(episode)">{{episode.title}}</button>
+              <button class="btn-link title" (click)="onChartSingleEpisode(episode)" [title]="episode.title">{{episode.title}}</button>
             </ng-template>
           </td>
           <td>{{episode.releaseDate}}</td>
           <td>{{episode.totalForPeriod | largeNumber}}</td>
-          <td>{{episode.avgPerIntervalForPeriod | largeNumber}}</td>
```
sandikbarr: #190 we made a decision to not include any averages at this time

src/app/ngrx/reducers/episode-metrics.reducer.ts
```diff
@@ -71,35 +71,29 @@ export function EpisodeMetricsReducer(state: EpisodeMetricsModel[] = initialStat
         return newState;
       }
       break;
-    case ACTIONS.ActionTypes.CASTLE_EPISODE_ALL_TIME_METRICS_SUCCESS:
-      if (action instanceof ACTIONS.CastleEpisodeAllTimeMetricsSuccessAction) {
-        const { id, seriesId } = action.payload.episode;
-        const epIdx = episodeIndex(state, id, seriesId);
+    // TODO: #141 & #142 should remove from this state and combine with performance metrics state
```
sandikbarr: When I'm able to create the selectors for the chart (and table) data and clean up those components, we'll get the total all time number from the performance metrics and drop it from this reducer.

src/app/shared/menu/dropdown.css
```diff
@@ -25,9 +25,9 @@
   white-space: nowrap;
 }
 .down-arrow {
-  border-left: 8px solid transparent;
-  border-right: 8px solid transparent;
-  border-top: 8px solid #676767;
+  border-left: 6px solid transparent;
```
sandikbarr: #201

src/app/ngrx/reducers/episode-performance-metrics.reducer.ts
```diff
@@ -0,0 +1,58 @@
...
+};
+
+export function EpisodePerformanceMetricsReducer(state: EpisodePerformanceMetricsState = initialState, action: ACTIONS.AllActions) {
+  switch (action.type) {
+    case ACTIONS.ActionTypes.CASTLE_EPISODE_PERFORMANCE_METRICS_LOAD: {
+      const {seriesId, id, guid} = action['payload'];
+      return {
+        entities: {
+          ...state.entities,
+          [id]: {seriesId, id, guid, loading: true, loaded: false}
```
sandikbarr: Now that we know whether these are still loading, we can add them to the loading state and display the spinner until they're done. #221

src/app/downloads/downloads-table.component.html
cavis: 👍

src/app/downloads/downloads-table.component.ts
```diff
@@ -41,7 +41,6 @@ export class DownloadsTableComponent implements OnDestroy {
     this.routerSub = this.store.select(selectRouter).subscribe((newRouterState: RouterModel) => {
       if (newRouterState) {
         if (isPodcastChanged(newRouterState, this.routerState)) {
-          this.store.dispatch(new ACTIONS.CastlePodcastAllTimeMetricsLoadAction());
```
cavis: Is this not needed anymore?

src/app/ngrx/effects/castle.effects.ts
```diff
@@ -72,57 +72,66 @@ export class CastleEffects {
...
-        // gonna be difficult to capture this error without changing shape of podcast metrics state since it's just an array
-        return Observable.of(new ACTIONS.CastlePodcastAllTimeMetricsFailureAction({podcast: undefined, error: 'No selected podcast yet'}));
-      }
+  loadPodcastPerformanceMetrics$: Observable<Action> = this.actions$
+    .ofType(ACTIONS.ActionTypes.CASTLE_PODCAST_PERFORMANCE_METRICS_LOAD)
+    .map((action: ACTIONS.CastlePodcastPerformanceMetricsLoadAction) => action.payload)
+    .switchMap((payload: ACTIONS.CastlePodcastPerformanceMetricsLoadPayload) => {
+    const { seriesId, feederId } = payload;
+      return this.castle
+        .followList('prx:podcast', {id: feederId})
```
cavis: I also noticed that castle and/or haldoc is misbehaving on this `prx:podcast` link ... should be a regular `follow()`, except that throws an error.  But you found something that does work, so :+4:!

src/app/ngrx/effects/castle.effects.ts
```diff
@@ -72,57 +72,66 @@ export class CastleEffects {
...
-                });
-              }
+    this.store.select(selectSelectedPodcast).subscribe((podcast: PodcastModel) => {
+      this.selectedPodcast = podcast;
+      if (this.selectedPodcast) {
+        const {seriesId, feederId} = this.selectedPodcast;
+        this.store.dispatch(new ACTIONS.CastlePodcastPerformanceMetricsLoadAction({seriesId, feederId}));
+      }
+    });
+    this.store.select(selectRecentEpisode).subscribe((episode: EpisodeModel) => {
```
cavis: Ooh, this is a good place for this.  Good call.

src/app/ngrx/reducers/episode-performance-metrics.reducer.ts
cavis: Is this firing twice?  When I reload the page, the metrics show, then disappear, then show up again.

http://cav.is/i/j7uct.gif

src/app/downloads/downloads-table.component.ts
sandikbarr: Nope, only needs to load podcast performance metrics in the castle effect when the selectSelectedPodcast selector fires. (Yeah, I'm seeing why ol' Todd prefixes those selector names with get instead of select.)

src/app/ngrx/effects/castle.effects.ts
sandikbarr: Credit to gcampo88 who set this up initially for the summary table

src/app/ngrx/reducers/episode-performance-metrics.reducer.ts
sandikbarr: I see what you mean about the flashing. The table component also loads the recent episode performance metrics, which on page 1 does load the same episode twice. So it updates the state once for the recent episode, then again for each of the episodes in the table. I _think_ this may be the cause of the flashing. It doesn't seem to do the flashing when the recent episode is not in the table (page > 1.)

sandikbarr:
The most recent episode performance metrics does need to be loaded both on page 1 for the table and for the profile component in case they start on a route other than page 1 downloads. Page 1 downloads should also refresh the data and update the profile component to keep values synced.

The flashing was due to loading the episode the second time. The loading action cleared the values. Fixed. Back to you @cavis!

cavis:
:+one:

Issue #228, opened 2018-02-05
clicking Downloads again when you're already on it uncharts Podcast


gcampo88:
steps:
1) load a show and go to the Stacked Downloads view. 
2) click Downloads again in left nav. The podcast becomes uncharted, and everything after `/{interval}` in the URL goes away

sandikbarr:
This got worse with the episode selection reset changes. Now it uncharts everything, but we have a helpful placeholder pointing to the Podcast chart. 🤓Bumping the priority to get this fixed sooner.

Issue #142, opened 2018-01-08
getMetricsSummaryData selector


sandikbarr:
a metrics summary data selector would combine routing state with podcast, episode, and metrics state to return data to be displayed in the summary table

sandikbarr:
Needs to be refactored to use RouteStateSerializer

pkarman:
can we bump this out since it is not user-facing?

sandikbarr:
I created the selectors when I added the summary info, but I'm gonna need to refactor it with this RouterStateSerializer PR, so it will be fully completed today.

sandikbarr:
This was already done

Issue #135, opened 2018-01-08
replace series dropdown with index page


sandikbarr:
We will remove the dropdown in the navbar, and top level navigation to different podcasts will be done via an index page that will eventually grow into our dashboard.

kookster:
We're not sure how much this will actually get used, so maybe build something simple but still useful.

Also need to consider what the menu looks like to link to this index page from on the page for a series (i.e. a link to "Series" in the nav, kinda like the nav links in publish?)

kookster:
we'll need to decide if this is pressing for the current sprint - and when we can get an index page designed and built before we can change/remove the dropdown

pkarman:
Let's bump this out of beta. @sandikbarr has recently added a scrolling nav dropdown to select from. We can revisit this after beta.

PR #243, opened 2018-02-14
redirect navigation back to '/' to previous router state

sandikbarr:


src/app/app.component.ts
```diff
@@ -30,8 +29,7 @@ export class AppComponent implements OnInit, OnDestroy {
 
   constructor(
     public store: Store<any>,
-    private angulartics2GoogleAnalytics: Angulartics2GoogleAnalytics,
-    private router: Router,
```
sandikbarr: extra, removed

src/app/ngrx/effects/routing.effects.ts
```diff
@@ -25,14 +25,26 @@ export class RoutingEffects {
...
     .map((action: ACTIONS.CustomRouterNavigationAction) => action.payload)
     .switchMap((payload: ACTIONS.CustomRouterNavigationPayload) => {
       const routerState: RouterModel = {...payload.routerState};
       // map to an action with our CUSTOM_ROUTER_NAVIGATION type
-      return Observable.of(new ACTIONS.CustomRouterNavigationAction({routerState}));
+      if (Object.keys(routerState).length > 0) {
+        // TODO: fix below and return Action rather than only sometimes dispatch
+        this.store.dispatch(new ACTIONS.CustomRouterNavigationAction({routerState}));
+      }
+      // TODO: #141 and #142
```
sandikbarr: This feels less good, but hopefully with all the TODOs I'm making on #141 and #142 I'll be able to re-address selectors soon and undo some weirdness

src/app/ngrx/effects/routing.effects.ts
```diff
@@ -156,6 +168,11 @@ export class RoutingEffects {
               private actions$: Actions) {
     this.store.select(selectRouter).subscribe(routerState => this.routerState = routerState);
     this.store.select(selectPodcasts).subscribe(podcasts => this.podcasts = podcasts);
+    this.router.events.filter(event => event instanceof RoutesRecognized).subscribe((event: RoutesRecognized) => {
```
sandikbarr: This is the big change. Catch router events to `'/'` and route back to the previous state. This does not give us a full
 cycle of router events on the route back to previous state though, hence the weirdness with customeRouterNavigation$ above.

pkarman: By "previous state" do you mean it acts like a Back button?

sandikbarr: Sort of. More like it backs out of actually routing to '/'. The app doesn't seem to actually change. The route appears to stay the same, and there is no subsequent router event because it matches the existing actual route.

sandikbarr:
I thought about changing the styleguide component with this home link to not be a link at all, but that felt like changing the styleguide for no good reason because apps should have a `/` route and the actual (hopefully near term) solution seems to be building this index page #135 

Downloads should be a routing action rather than a `[routerLink]` https://github.com/PRX/metrics.prx.org/issues/228. It's dropping the parameters from the link.

PR #244, opened 2018-02-14
Pagination toolbar styles

pkarman:
* Rely on the styleguide for pager styles
* Change text of the show/hide toggle for episode table
* Add first/last page pagination buttons
* Pagination has a "sliding window" currently configured to 5 page links. Logic from http://jasonwatmore.com/post/2016/01/31/angularjs-pagination-example-with-logic-like-google

DO NOT MERGE till https://github.com/PRX/styleguide.prx.org/pull/90 is deployed.

### DIFF
--- a/src/app/downloads/downloads-table.component.css
+++ b/src/app/downloads/downloads-table.component.css
@@ -55,7 +55,6 @@ table.scroll-x th:first-child, table.scroll-x td:first-child {
   display: flex;
   justify-content: space-between;
   padding: 3px;
-  background-color: #fff;
   margin-top: 3px;
 }
 .table-navigation > button {
--- a/src/app/downloads/downloads-table.component.css
+++ b/src/app/downloads/downloads-table.component.css
@@ -54,6 +54,9 @@ table.scroll-x th:first-child, table.scroll-x td:first-child {
 .table-navigation {
   display: flex;
   justify-content: space-between;
+  padding: 3px;
+  background-color: #fff;
+  margin-top: 3px;
 }
 .table-navigation > button {
   text-align: left;
--- a/src/app/downloads/downloads-table.component.html
+++ b/src/app/downloads/downloads-table.component.html
@@ -3,8 +3,8 @@
 
   <div class="table-navigation">
     <button (click)="toggleExpandedReport()" class="btn-link">
-      <p *ngIf="!expanded">Show more &gt;&gt;</p>
-      <p *ngIf="expanded">Show less &lt;&lt;</p>
+      <p *ngIf="!expanded">Show daily numbers &#187;</p>
+      <p *ngIf="expanded">Hide daily numbers &#171;</p>
     </button>
 
     <metrics-episode-page
--- a/src/app/downloads/downloads-table.component.html
+++ b/src/app/downloads/downloads-table.component.html
@@ -3,8 +3,8 @@
 
   <div class="table-navigation">
     <button (click)="toggleExpandedReport()" class="btn-link">
-      <p *ngIf="!expanded">Show daily numbers &#187;</p>
-      <p *ngIf="expanded">Hide daily numbers &#171;</p>
+      <p *ngIf="!expanded">Show {{routerState.interval.key}} numbers &#187;</p>
+      <p *ngIf="expanded">Hide {{routerState.interval.key}} numbers &#171;</p>
     </button>
 
     <metrics-episode-page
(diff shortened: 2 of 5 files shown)


pkarman:
screen shot 2018-02-13 at 7 04 59 pm

src/app/downloads/downloads-table.component.html
```diff
@@ -3,8 +3,8 @@
 
   <div class="table-navigation">
     <button (click)="toggleExpandedReport()" class="btn-link">
-      <p *ngIf="!expanded">Show more &gt;&gt;</p>
-      <p *ngIf="expanded">Show less &lt;&lt;</p>
+      <p *ngIf="!expanded">Show daily numbers &#187;</p>
```
gcampo88: I think you'll want to use the key from the interval here -- the extra info included in the table will only be `daily` if the user has selected Daily interval

src/app/shared/paging/episode-page.component.ts
```diff
@@ -20,17 +22,36 @@ export class EpisodePageComponent implements OnChanges {
   @Input() totalPages;
   @Output() pageChange = new EventEmitter<number>();
   pages: number[];
-  showNumPages = 10;
+  showNumPages = 5;
```
gcampo88: there's a test you'll need to update for this one -- currently failing 
`should only show up to showNumPages at a time` => 	`Expected 5 to equal 10.`

src/app/downloads/downloads-table.component.html
sandikbarr: yep, yep, `routerState.interval.key` (mostly making this comment to say I don't wanna talk about why the interval names are the same as the key, things morph into other things ok. 😺)

src/app/downloads/downloads-table.component.css
```diff
@@ -54,6 +54,9 @@ table.scroll-x th:first-child, table.scroll-x td:first-child {
 .table-navigation {
   display: flex;
   justify-content: space-between;
+  padding: 3px;
+  background-color: #fff;
+  margin-top: 3px;
```
sandikbarr: This is very much just a personal preference, probably should get @farski's opinion as well. I'm not feeling the white background in the table-navigation bar so much. What if the buttons where white with no background on that div?
Instead of:
screen shot 2018-02-14 at 9 06 29 am
we could do this:
screen shot 2018-02-14 at 9 07 46 am

pkarman: It seems like the light grey is reserved for background everywhere else, and that anything clickable is on a white background. It felt "unfinished" somehow to me in that regard. But I'm happy to defer to you and @farski

src/app/shared/paging/episode-page.component.ts
pkarman: good catch. seems like the test should be?
```diff
-    expect(comp.pager.pagesEnd).toEqual(10);
+    expect(comp.pager.pagesEnd).toEqual(comp.pager.showNumPages);
```
to prevent that from happening if the config ever changes. Do we want to test config specifics, or just that the config is respected in the code execution?

src/app/downloads/downloads-table.component.css
```diff
@@ -54,6 +54,9 @@ table.scroll-x th:first-child, table.scroll-x td:first-child {
 .table-navigation {
   display: flex;
   justify-content: space-between;
+  padding: 3px;
+  background-color: #fff;
```
sandikbarr: 2 votes for remove the background
https://prx.slack.com/archives/G82E05W2G/p1518707335000076

If the buttons stay white, they aren't gonna look like ghost buttons as much, but they aren't btn-link either.

src/app/downloads/downloads-table.component.css
pkarman: background reverted to grey

screen shot 2018-02-15 at 12 30 43 pm

src/app/shared/paging/episode-page.component.ts
```diff
@@ -20,24 +31,45 @@ export class EpisodePageComponent implements OnChanges {
...
+      }
+      else {
+        this.pagesBegin = this.currentPage - halfWindow;
+        this.pagesEnd = this.currentPage + halfWindow;
+      }
+    }
+
     this.pages = [];
-    for (let i = 1; i <= this.totalPages; i++) {
+    for (let i = this.pagesBegin; i <= this.pagesEnd; i++) {
```
pkarman: instead of an array of all pages and then slicing in the html section, create an array of only the visible page numbers and a simple for loop in the html.

src/app/shared/paging/episode-page.component.ts
sandikbarr: testing that the config is respected is my preference

PR #248, opened 2018-02-15
use minY=0 on episodes chart

sandikbarr:
Except for the multiline chart where we were sometimes seeing a negative y axis when there was a lot of zeroes, it seemed best to let C3 make the decisions about the y axis.


pkarman:
heh spoke too soon - failing test?

sandikbarr:
Weird awscli error. Phew. Retry fixed it!

PR #249, opened 2018-02-16
Fix clicking downloads link when you're already on it

sandikbarr:


src/app/shared/menu/date/custom-date-range-dropdown.component.css
```diff
@@ -44,6 +44,15 @@
   min-width: 285px;
 }
 
+@media (max-width: 768px) {
```
sandikbarr: This menu doesn't fit in small screen sizes, so I decided to not display the redundant controls for the list of standard ranges until we can prioritize responsive and mobile design. :deal-with-it-parrot:

pkarman: confirmed, looks ok

screen shot 2018-02-16 at 11 14 25 am

sandikbarr: 👍 Yep, much better as a temporary measure at least. Google Analytics reports almost 20% mobile users.

Issue #137, opened 2018-01-08
move default filter to routing effects


sandikbarr:
When first loading the default filter is determined for date range and interval. This currently happens in the downloads component but should be incorporated with filterFromRoute$ in the routing.effects.

sandikbarr:
This will also fix a bug where if the page is refresh on demographics or traffic sources routes the sidebar nav doesn't have any links because Ngrx doesn't have the series id.

pkarman:
@sandikbarr is this bug still present? I didn't notice it during your demo.

sandikbarr:
It is, but I didn't refresh on Demographics or Traffic Sources routes. I did however get it fixed with my router state serializer, the PR I'm currently working on. So yes, we can bump this. (Even though I'm very close to making this change in the refactor.)

PR #254, opened 2018-03-06
downloads chart and table selectors

sandikbarr:
The chart and download component setup is all done with composed selectors now, much better. The Downloads Component loading setup still needs a bit of help (#137, etc) with how metrics data is loaded.


src/app/downloads/downloads.component.spec.ts
```diff
@@ -6,75 +6,43 @@ import { StoreModule, Store } from '@ngrx/store';
...
-    ['2017-08-31T00:00:00Z', 162900],
-    ['2017-09-01T00:00:00Z', 46858],
-    ['2017-09-02T00:00:00Z', 52522],
-    ['2017-09-03T00:00:00Z', 162900],
-    ['2017-09-04T00:00:00Z', 46858],
-    ['2017-09-05T00:00:00Z', 52522],
-    ['2017-09-06T00:00:00Z', 162900],
-    ['2017-09-07T00:00:00Z', 46858]
-  ];
+  const metricsPropertyName = getMetricsProperty(routerState.interval, routerState.metricsType);
```
sandikbarr: Side note: I hate myself a little bit for letting that property name leak outside the reducer; should have used two separate params for the action.

src/app/ngrx/reducers/models/downloads-table.model.ts
```diff
@@ -0,0 +1,12 @@
+import { TimeseriesDatumModel } from 'ngx-prx-styleguide';
+
+export interface DownloadsTableModel {
```
sandikbarr: This one is actually a view model, not a state model. I don't love having it with the reducer models... It was a last minute change to resolve a TypeScript warning where one of the imports in a component wasn't able to see it in the selector file.

src/app/ngrx/reducers/router.serializer.ts
```diff
@@ -75,7 +75,7 @@ export class CustomSerializer implements RouterStateSerializer<RouterModel> {
         router.endDate = endDate;
       }
       if (params['episodes']) {
-        router.episodeIds = params['episodes'].split(',').map(stringValue => +stringValue);
+        router.episodeIds = params['episodes'].split(',').map(stringValue => +stringValue).filter(id => id > 0);
```
sandikbarr: Empty string was resulting in a single entry of zero, episode Ids are > 0. Saw a bug, so made a quick fix.

src/app/shared/util/metrics.util.ts
```diff
@@ -1,23 +1,8 @@
-import { PodcastModel, EpisodeModel, RouterModel,
+import { EpisodeModel, RouterModel,
   PodcastMetricsModel, EpisodeMetricsModel, getMetricsProperty,
   IntervalModel, INTERVAL_HOURLY, INTERVAL_DAILY, INTERVAL_WEEKLY, INTERVAL_MONTHLY } from '../../ngrx';
 import * as dateUtil from './date/date.util';
 
-export const filterPodcasts = (filter: RouterModel, podcasts: PodcastModel[]): PodcastModel => {
```
sandikbarr: This filtering happens in the selectors now.

src/testing/downloads.fixtures.ts
```diff
@@ -0,0 +1,117 @@
+import { RouterModel, EpisodeModel, PodcastModel, ChartType, MetricsType,
+  CHARTTYPE_STACKED, INTERVAL_DAILY, METRICSTYPE_DOWNLOADS } from '../app/ngrx';
+import { MockHalDoc } from 'ngx-prx-styleguide';
+
+const podcast: PodcastModel = {
```
sandikbarr: Could probably be providing better fixture data, but this has needed to happen for a while.

src/app/ngrx/reducers/episode-metrics.reducer.ts
```diff
@@ -5,12 +5,10 @@ export interface EpisodeMetricsModel {
   id: number;
   guid?: string;
   page?: number;
-  charted?: boolean;
```
cavis: So we're not tracking the "charted" state anymore?

src/app/shared/util/metrics.util.ts
cavis: 👍

src/testing/downloads.fixtures.ts
cavis: Oooh - nice.

src/app/downloads/downloads-table-presentation.component.html
```diff
@@ -24,14 +24,14 @@
...
-        <tr>
+        <tr *ngIf="podcastTableData">
           <td>
             <prx-checkbox *ngIf="showPodcastToggle; else podcastTitle"
               small [checked]="podcastTableData.charted" [color]="podcastTableData.color"
-              (change)="toggleChartPodcast($event)">{{podcastTableData.title}}</prx-checkbox>
+              (change)="toggleChartPodcast.emit($event)">{{podcastTableData.title}}</prx-checkbox>
             <ng-template #podcastTitle><span class="title">{{podcastTableData.title}}</span></ng-template>
           </td>
-          <td>{{podcastTableData.releaseDate}}</td>
```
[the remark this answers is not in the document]
sandikbarr: yep, I just didn't put it on the model for this update. For episodes, I'm formatting the publishedAt field from the template rather than having an already formatted field sitting on the model.

src/app/ngrx/effects/routing.effects.ts
```diff
@@ -25,26 +25,14 @@ export class RoutingEffects {
   // The router reducer captures the CustomRouterNavigationAction to save routing state.
   // The CustomRouterNavigationAction is mocked by tests but is _not_ otherwise ever manually called by the application,
   // only through routing and the router-store connecting module.
-  @Effect({dispatch: false})
-  customRouterNavigation$: Observable<void> = this.actions$
+  @Effect()
```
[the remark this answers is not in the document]
sandikbarr: Yes, this always returns an action now. When I had charted on the metrics data model and added that work around for the Home link, an empty routing action was getting through, and it cleared out charted on the episode metrics. Having charted on the episode metrics was a poor decision on my part. I stuck charted into the metrics where I thought we'd use it from because I was struggling with selector composition. (For some reason when I first tried it, the selectors would fire updates constantly instead of just when there were changes.) It was not a good place to keep that state. With selector composition set up correctly, I've now been able to pull it back out and only keep charted in the router state.

src/app/ngrx/reducers/episode-metrics.reducer.ts
sandikbarr: It's only on the router state now and is using selector composition to combine it with metrics for the chart data selector

src/app/ngrx/reducers/episode-metrics.reducer.ts
```diff
@@ -5,12 +5,10 @@ export interface EpisodeMetricsModel {
...
   monthlyDownloads?: any[][];
   weeklyDownloads?: any[][];
   dailyDownloads?: any[][];
   hourlyDownloads?: any[][];
-  allTimeDownloads?: number;
```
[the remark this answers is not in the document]
sandikbarr: I added performance metrics reducers when I put that data into the profile section. All time downloads was still in here from when we added it to use in the summary table, so the data was used in both places for a time (which made me pretty sad so I made a lot of TODO comments about fixing it.) With this refactor, I've been able to remove it from this reducer and combine performance metrics with episode and metrics data in the selectors. So now we're only keeping state in one place, how it should have been in the first place, but you know, things happen.

PR #258, opened 2018-03-26
reconcile all time total and recent episode today values

sandikbarr:
- When the total sum of metrics data values is larger than the all time totals from the rollup "performance" metrics endpoint, use the total value from the metrics data
- Also, use metrics data for recent episode today value if available in state and larger than what came from the rollup "performance" metrics


src/app/ngrx/reducers/selectors/downloads-table.selectors.spec.ts
```diff
@@ -54,8 +55,13 @@ describe('Downloads Table Selectors', () => {
       expect(result.title).toEqual('All Episodes');
     });
 
-    it('should include podcast all time total downloads', () => {
-      expect(result.allTimeDownloads).toEqual(podPerformance.total);
+    it('should include podcast all time total downloads, but should use total for range if larger', () => {
+      if (metricsUtil.getTotal(podDownloads) > podPerformance.total) {
```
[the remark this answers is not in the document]
cavis: I guess i was expecting a test for each case, rather than the if/else within 1 test.

src/app/ngrx/reducers/selectors/recent-episode.selectors.ts
```diff
@@ -13,3 +18,40 @@ export const selectRecentEpisode = createSelector(selectRecentEpisodeEntities, s
...
+        // if we have full metrics data on the recent episode, check the today and all time total values and reconcile
+        let totalForToday = entities[episode.id].today;
+        // if recent episode is not on routed episode page, check for episode dailyDownloads and total
+        const totalForRange = getTotal(metricsData(routerState, metrics)) || (metrics.dailyDownloads && getTotal(metrics.dailyDownloads));
+
+        // today's total doesn't update until day after release, so attempt to get it from the daily downloads
+        // (If app was loaded from non default url, we probably won't have dailyDownloads for today, an edge case I can live with.
+        // _could_ also total hourlies from today if they are refreshing the page from that route, but this is not doing that.)
+        if (metrics.dailyDownloads &&
+          new Date(metrics.dailyDownloads[metrics.dailyDownloads.length - 1][0]).valueOf() >= beginningOfTodayUTC().valueOf()) {
```
cavis: Do you need to check the date for this?  Or could you just always pick the larger number (and default to 0).  Date comparisons just always make me nervous.  And who knows - maybe Castle will get fixed someday?

cavis: Or I could just be misreading this.

src/app/ngrx/reducers/selectors/downloads-table.selectors.spec.ts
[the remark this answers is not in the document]
sandikbarr: It's fixture data, so yeah only one of these things is getting tested I suppose. I'll add some additional data that will fit both cases and split out the test.

src/app/ngrx/reducers/selectors/recent-episode.selectors.ts
sandikbarr: yeah, I it does need to check the date because if you loaded the app from a non default route that doesn't include today's date but does include the most recent episode, the last data point wouldn't be the today value. We probably will want to fix Castle eventually as I've heard that day 0, 7 days, and 30 days (but by 30 we really mean 28) are the important rollups for episodes, so we'll likely want those on a dashboard.

sandikbarr: (they could load the app on a non default route if say for example they refreshed a page with yesterday's url where the end date is no longer the current day)

sandikbarr:
Thanks for the feedback! Back to you when you get a chance. Thanks again!

src/app/ngrx/reducers/selectors/recent-episode.selectors.ts
cavis: oh, right.  makes good sense then!

PR #265, opened 2018-04-30
all episodes was disappearing from table when toggled off

sandikbarr:


src/app/ngrx/reducers/selectors/downloads-table.selectors.ts
```diff
@@ -23,8 +24,7 @@ export const selectDownloadTablePodcastMetrics = createSelector(
    podcastPerformanceMetrics: PodcastPerformanceMetricsModel): DownloadsTableModel => {
     let podcastData: DownloadsTableModel;
 
-    if (podcastMetrics &&
-      routerState.chartType === CHARTTYPE_PODCAST || (routerState.chartPodcast && routerState.chartType === CHARTTYPE_STACKED)) {
```
sandikbarr: Not sure what I was thinking here, oops. Probably just a dumb mistake because I was making a lot of similar changes to refactor for these selectors at once.

PR #278, opened 2018-06-01
update header/footer

sandikbarr:
goes with https://github.com/PRX/styleguide.prx.org/pull/111

a couple more updates here to go with the header update
screen shot 2018-06-04 at 4 59 11 pm

### DIFF
--- a/src/app/app.component.spec.ts
+++ b/src/app/app.component.spec.ts
@@ -29,7 +29,7 @@ describe('AppComponent', () => {
   let el: HTMLElement;
   let store: Store<any>;
   const userinfo = new Userinfo();
-  userinfo.name = 'Joey JoJo Jr Shabadoo';
+  userinfo.email = 'Joey@Shabadoo.pizza';
 
   const account: AccountModel = {id: 1234, name: 'Joey JoJo Jr Shabadoo'};
   const podcasts: PodcastModel[] = [{seriesId: 9876, title: 'Foobar'}];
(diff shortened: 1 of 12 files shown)


src/app/app.component.spec.ts
```diff
@@ -29,7 +29,7 @@ describe('AppComponent', () => {
   let el: HTMLElement;
   let store: Store<any>;
   const userinfo = new Userinfo();
-  userinfo.name = 'Joey JoJo Jr Shabadoo';
+  userinfo['name'] = userinfo['email'] = 'Joey@Shabadoo.pizza';
```
sandikbarr: (kept `userinfo.name` in this test so it would pass while I'm still waiting on styleguide update to be published)

src/assets/zendesk-widget.js
```diff
@@ -13,3 +13,13 @@ window.zEmbed || function (e, t) {
     n && (this.domain = n), e.id = "js-iframe-async", e.src = "https://assets.zendesk.com/embeddable_framework/main.js", this.t = +new Date, this.zendeskHost = "prx.zendesk.com", this.zEQueue = a, this.body.appendChild(e)
   }, o.write('<body onload="document._l();">'), o.close()
 }();
+
+window.zESettings = {
+  webWidget: {
+    // https://developer.zendesk.com/embeddables/docs/widget/api
+    // is possible to set positioning and offset to top right but cannot configure button padding
```
sandikbarr: I messed around with putting this in the header per the mockup, but I found that you cannot configure the padding on the button. I found that I had to change the button position with the changing size of the header with those responsive settings. Our media query didn't match up to what they were doing to determine mobile layouts, so I couldn't use offset to handle this positioning. It all just ended up feeling like a bad idea.

cavis: I'm also just noticing that we don't have the feedback widget in Publish.  Feels like it should be there, right?

Issue #170, opened 2018-01-19
responsive menu/controls


sandikbarr:
implement #169
Mobile:[Mobile.pdf](https://github.com/PRX/metrics.prx.org/files/2051196/Mobile.pdf)

Desktop: 
![desktop](https://user-images.githubusercontent.com/2250413/40687753-495e97f8-6361-11e8-8896-a2f79eeaef85.png)

* [ ] the advanced/custom date range picker menu is not shown for screen (or window?) sizes below ___ (as is the menu is 900px wide)
* [ ] the menu controls are no longer a distinct bar of buttons/dropdowns and instead are left/right aligned dropdowns without borders
* [ ] in desktop mode, the menus have an `#FF9600` orange vertical bar separator but no other distinct borders
* [ ] Is that dropdown arrow in desktop mode for Operating System or for the date range picker? Is the separate quick selector date range picker no longer shown in desktop mode, only in the full advanced menu?
* [ ] the menu items have swapped location, interval/bucket is to the right and date picker to the left of the interval/bucket dropdown <-- I actually couldn't do this because that advanced date range menu is so large even in desktop mode
* [ ] Has the summary information 'IN THIS RANGE' moved into the menu heading?
* [ ] How does the user select different chart types? Did those buttons move somewhere else? Mobile says that the user will swipe left/right though different chart types, but what about desktop? Do we want to expose any buttons for this on mobile or swipe only? (asked Farski about swiping, he thought there wouldn't be room for buttons, but there is so swiping has not been implemented at this time)
* [ ] in mobile mode, the PRX logo moves to the far right of the header and the show dropdown selector moves to the far left. Where does the user image and app menu go (is this because mobile isn't supported in exchange or publish)?

Issue #279, opened 2018-06-08
Tables


farski:
- Work for various number of columns/data density
- Some sortable columns
- Paging
- Responsive
- Item selection

farski:
For very basic tables, meaning those with a single primary value and and perhaps some associated or derived value, the styling is stark and simple

![image](https://user-images.githubusercontent.com/3483/41560299-b454ad64-7314-11e8-97a9-c7e7107cb4e9.png)

Small formats have only a single value per column, and the sizing and spacing should collapse along with the screen real estate. There's no strict number of breakpoints, two or three for all small format is probably enough. There should always be enough space for the value columns and their titles, without feeling too cramped. If something needs to get truncated it should be the label of the dimension value.

At larger/desktop sizes, the value and a derived value would collapse to a single column (that's mainly to prevent cases on very wide screens from having the data spread way out in multiple columns. When it's possible to only have two columns in this way, the first (name) column width should be capped, and the second column can expand as necessary to fill the space).

On desktop some columns can be made sortable, so there's an indicator.

In most cases one column of data in the table corresponds to the data that is being displayed the chart; that column should be highlighted.

![image](https://user-images.githubusercontent.com/3483/41560563-6ce78928-7315-11e8-8d66-300d60e5442b.png)

farski:
For some data, there is additional data nested below rows displayed by default in the table (like in the case of the Countries table, where some countries have other geographic subregions).

On very small screens expanding the nested data is not supported.

On most modern mobile sizes, the rows can expand to show the next level of data. Only one nesting can be visible at a time (expanding a second collapses the first). Percentage values aren't displayed for nested data (since it'd confusing to explain what that's a percent of)

![image](https://user-images.githubusercontent.com/3483/41561693-9965e190-7318-11e8-8e3e-027674884fcf.png)

Desktop formats follow the same nesting behavior with the above desktop design. The disclosure controls are more obvious on the desktop, since there's more room to play with and it's a more obvious/useful interaction on larger screens.

![image](https://user-images.githubusercontent.com/3483/41561836-f491ff18-7318-11e8-8d2f-951b86648ff0.png)

farski:
For tables with more data/columns, the format on desktop follows pretty much the same pattern, just with more columns! The charted column should still be highlighted.

![image](https://user-images.githubusercontent.com/3483/41562370-58cdb8d6-731a-11e8-884d-43f791215a12.png)

[...]

PR #285, opened 2018-06-19
responsive sidenav

sandikbarr:
implements responsive side nav #285 and responsive menu and controls #170 

~TODO: NEEDS STYLEGUIDE UPDATE TO INCLUDE ICONS~ ✅ done


src/app/downloads/downloads.component.ts
```diff
@@ -96,8 +96,9 @@ export class DownloadsComponent implements OnInit, OnDestroy {
 
             if (this.updatePodcast && this.routerState.podcastSeriesId) {
               this.podcast = this.podcasts.find(p => p.seriesId === this.routerState.podcastSeriesId);
-              this.totalPages = this.podcast.doc.count('prx:stories') / EPISODE_PAGE_SIZE;
-              if (this.podcast.doc.count('prx:stories') % EPISODE_PAGE_SIZE > 0) {
+              const numEpisodes = this.podcast ? this.podcast.doc.count('prx:stories') : 0;
```
sandikbarr: stumbled on this bug with a podcast route that my newly logged in user was not authorized for

src/app/shared/menu/date/custom-date-range-dropdown.component.ts
```diff
@@ -46,11 +46,15 @@ import { GoogleAnalyticsEventAction, RouteAdvancedRangeAction } from '../../../n
 export class CustomDateRangeDropdownComponent {
   @Input() routerState: RouterModel;
   tempRange: RouterModel;
-  open = false;
   intervalList = IntervalList;
   userChoseRange: string;
   CUSTOM_DATE = 'custom-date';
   STANDARD_DATE = 'standard-date';
+  open = false;
```
sandikbarr: noticed especially in mobile mode when there is more content to scroll that these dropdowns remained open but were out of place when the window scrolled

src/app/shared/menu/date/standard-date-range-dropdown.component.ts
```diff
@@ -32,8 +25,11 @@ import * as dateUtil from '../../util/date';
 export class StandardDateRangeDropdownComponent {
   @Input() standardRange: string;
   @Input() interval: IntervalModel;
-  @Output() custom = new EventEmitter();
   open = false;
+  @HostListener('window: scroll', [])
+  onWindowScroll() {
+    this.open = false;
```
cavis: Would it be good to also do this on the other dropdowns?

cavis: Ah, i see it's on the canned interval dropdown.  But maybe on the custom one too?

src/app/shared/nav/nav-menu.component.ts
```diff
@@ -1,50 +1,52 @@
...
-    </nav>
+    <div class="dropdown" [class.open]="open">
+      <div class="overlay" (click)="toggleOpen()"></div>
+      <div class="dropdown-button">
+        <button (click)="toggleOpen()" >{{routerState.metricsType}}<span class="down-arrow"></span></button>
+      </div>
+      <div class="dropdown-content rollout">
+        <nav>
+          <li *ngFor="let t of types">
+            <button
```
cavis: In mobile view ... when i click/select on these buttons, the dropdown doesn't close.  Have to click outside of the dropdown to make it close.

src/app/shared/podcast-nav/podcast-nav-dropdown.component.ts
```diff
@@ -32,7 +32,7 @@ export class PodcastNavDropdownComponent {
   }
 
   onPodcastChange(val: PodcastModel) {
-    if (val && val.seriesId !== this.selectedPodcast.seriesId) {
+    if (val && (!this.selectedPodcast || val.seriesId !== this.selectedPodcast.seriesId)) {
```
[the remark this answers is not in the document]
cavis: +1.  i think i've hit this before, when masquerading as different users.

src/app/shared/menu/date/custom-date-range-dropdown.component.ts
```diff
@@ -10,7 +10,10 @@ import { GoogleAnalyticsEventAction, RouteAdvancedRangeAction } from '../../../n
     <div class="dropdown" [class.open]="open">
       <div class="overlay" (click)="toggleOpen()"></div>
       <div class="dropdown-button">
-        <button class="btn-icon icon-calendar grey-dove" (click)="toggleOpen()" aria-label="Custom Date Range"></button>
+        <button (click)="toggleOpen()">
+          <metrics-date-range-summary [routerState]="routerState"></metrics-date-range-summary>
```
cavis: I'm noting this component has Hourly/Daily/Weekly/Monthly buttons in it.  Are those necessary, since the dropdown for those is right next to this button?

src/app/downloads/downloads.component.ts
```diff
@@ -16,15 +16,14 @@ import { isPodcastChanged, isBeginDateChanged, isEndDateChanged, isIntervalChang
   template: `
     <prx-spinner *ngIf="loading$ | async" overlay="true"
                  loadingMessage="Please wait..."></prx-spinner>
-    <router-outlet name="sidenav"></router-outlet>
-    <section class="content" *ngIf="loaded$ | async">
+    <section *ngIf="loaded$ | async">
       <metrics-menu-bar></metrics-menu-bar>
       <metrics-downloads-chart></metrics-downloads-chart>
       <metrics-downloads-table [totalPages]="totalPages"></metrics-downloads-table>
```
cavis: I'm noting that, in mobile, the table + pager get a little funky.  (Horizontal scrolling, page-number-overflow, etc).  Will that be addressed in a different ticket/PR?

src/app/shared/menu/date/standard-date-range-dropdown.component.ts
sandikbarr: I did actually have it on the custom date range dropdown but I took it out. I found that at my screen resolution, the window needed to be almost as tall as the viewport to see the Apply button. So if the user tried to scroll to it, the dropdown closes, and everyone is sad. I added these because it was gross to me that the dropdowns were fixed on scroll, but not being able to use the is kind of worse,

src/app/shared/menu/date/custom-date-range-dropdown.component.ts
sandikbarr: We didn't have any changes to the content of the desktop dropdowns in the designs, so I'm not sure if we meant to keep the quick interval selector and have the advanced all option selector on desktop, or what, but this is how I interpreted it.

src/app/downloads/downloads.component.ts
sandikbarr: There is a separate set of tickets to update the table.

src/app/shared/nav/nav-menu.component.ts
sandikbarr: will take a look!

src/app/downloads/downloads.component.ts
sandikbarr: tables design is here https://github.com/PRX/metrics.prx.org/issues/279

sandikbarr:
On the subject of page height, I message around with vh units on the `article` element. We don't have a defined height on the footer, so we can't really `calc(100vh - XZYpx)` but maybe we could do something different there. Without including height for the footer, it looked especially odd to me on tall mobile screens to push the footer down below a big empty COMING SOON screen.

I think we can continue making improvements on things like button size and other thoughts that come up. The purpose of this PR is to add responsive support to the menu bar (now a less defined bar but those same date range and interval controls) and to the side navigation, but I'm sure the mobile friendly conversation definitely isn't over.

I also liked the always visible series image. If we bring it back, I hope we can talk about why it was circular instead of square since podcast cover images are designed to be square.

Other than that, hopefully I've addressed your comments @cavis. Thanks!

pkarman:
lgtm!
screen shot 2018-06-26 at 9 52 17 am

PR #300, opened 2018-08-20
Fix for dup episode titles and episode metrics "loaded"

sandikbarr:
* Fixes the issue with duplicate episode titles (Showcase is a good example.) The chart maps the labels into an object, so the labels need to be unique.
* Fixes another issue with "loaded" when flipping between pages on the episode chart. There is a brief flash of a placeholder from components that should not display until data is loaded. In order to make "loading" stick while other data loading is in progress, set the episode metrics loaded to false when an episode page is loaded. Expecting that the download data will be need to be loaded into the state before displaying any components trying to use it.

### DIFF
--- a/src/app/ngrx/reducers/episode-metrics.reducer.ts
+++ b/src/app/ngrx/reducers/episode-metrics.reducer.ts
@@ -22,6 +22,21 @@ const episodeIndex = (state: EpisodeMetricsModel[], guid: string) => {
 
 export function EpisodeMetricsReducer(state: EpisodeMetricsModel[] = initialState, action: ACTIONS.AllActions) {
   switch (action.type) {
+    case ACTIONS.ActionTypes.CASTLE_EPISODE_PAGE_SUCCESS: {
+      // sets loaded to false on entry when episode page is loaded expecting that metrics will also be loaded
+      // (prevents loaded selector from showing prematurely when metrics load call has momentarily not yet been made)
+      const {episodes, page} = action.payload;
+      let newState = state;
+      episodes.forEach(episode => {
+        const {guid, podcastId} = episode;
+        const epIdx = episodeIndex(state, guid);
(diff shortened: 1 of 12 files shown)


src/app/ngrx/reducers/episode-metrics.reducer.ts
```diff
@@ -22,6 +22,21 @@ const episodeIndex = (state: EpisodeMetricsModel[], guid: string) => {
...
+      // sets loaded to false on entry when episode page is loaded expecting that metrics will also be loaded
+      // (prevents loaded selector from showing prematurely when metrics load call has momentarily not yet been made)
+      const {episodes, page} = action.payload;
+      let newState = state;
+      episodes.forEach(episode => {
+        const {guid, podcastId} = episode;
+        const epIdx = episodeIndex(state, guid);
+        if (epIdx === -1) {
+          episode = {podcastId, guid, page, loaded: false};
+          newState = [episode, ...state.slice(epIdx + 1)];
```
cavis: Isn't that always going to be `state.slice(0)`?

src/app/ngrx/reducers/podcast.reducer.ts
```diff
@@ -25,7 +25,12 @@ export function reducer(
 ): State {
   switch (action.type) {
     case ActionTypes.CASTLE_PODCAST_PAGE_SUCCESS: {
-      return adapter.addMany(action.payload.podcasts, state);
+      return adapter.upsertMany(action.payload.podcasts.map(podcast => {
```
cavis: Upsert ... neat!

src/app/ngrx/reducers/episode-metrics.reducer.ts
sandikbarr: related: I want to update these reducers to use @ngrx/entity also

PR #302, opened 2018-08-29
user agents for the whole podcast

sandikbarr:
user agents round 1, just shows data on the podcast, not yet on individual episodes.
also to be added: growth column in table


src/app/geo/soon.component.ts
```diff
@@ -0,0 +1,23 @@
+import { Component, Input } from '@angular/core';
+import {GROUPTYPE_GEOCOUNTRY, GROUPTYPE_GEOMETRO, GROUPTYPE_GEOSUBDIV, RouterParams} from '../ngrx';
+
+@Component({
+  selector: 'metrics-soon',
+  template: `
+    <section class="soon"
```
cavis: Nice.

src/app/ngrx/actions/cms.action.creator.ts
```diff
@@ -21,3 +21,12 @@ export class CmsAccountFailureAction implements Action {
   readonly type = ActionTypes.CMS_ACCOUNT_FAILURE;
   constructor(public payload: any) {}
 }
+
+export interface CmsAccountRetryActionsPayload {
```
cavis: Hmm... I'm confused what this is.  I thought all the logic to display the user login was in Styleguide now - and that it hits `id.prx.org/userinfo` instead of the auth/accounts endpoint in CMS.

cavis: In fact, I do see the ajax firing to get id-userinfo.  But I also see the requests going to CMS.

src/app/ngrx/effects/cms.effects.ts
```diff
@@ -41,7 +43,36 @@ export class CmsEffects {
     })
   );
 
+  @Effect()
+  retryActions$: Observable<Action> = this.actions$.pipe(
+    ofType(ACTIONS.ActionTypes.CMS_ACCOUNT_RETRY_ACTIONS),
+    filter((action: ACTIONS.CmsAccountRetryActionsAction) => action.payload.actions && action.payload.actions.length > 0),
+    map((action: ACTIONS.CmsAccountRetryActionsAction) => action.payload),
+    mergeMap((payload: ACTIONS.CmsAccountRetryActionsPayload) => {
+      return this.auth.token.pipe(
```
cavis: Really feels like this shouldn't be necessary here ... if token refreshing isn't working, it's a problem [in the HalDoc lib](https://github.com/PRX/styleguide.prx.org/blob/master/src/lib/src/hal/remote/halremote.ts#L133).  Which should just seamlessly retry the last request.  I guess I'd rather try to track down the failure there, if possible.  Checking `this.auth` and parsing tokens is a bit off-topic for this app.

src/app/ngrx/effects/cms.effects.ts
```diff
@@ -41,7 +43,36 @@ export class CmsEffects {
...
+        first(),
+        mergeMap(token => {
+          if (token) {
+            if (!this.auth.parseToken(token)) {
+              return Observable.of(new ACTIONS.CmsAccountFailureAction({error: 'Permission denied'}));
+            } else {
+              payload.actions.forEach(action => this.store.dispatch(action));
+              return this.cms.individualAccount.map(doc => {
```
cavis: Yeah, this is what i don't think you need.  The `prx-navuser` in the styleguide now gets its info [from the auth user-info service](https://github.com/PRX/styleguide.prx.org/blob/master/src/lib/src/header/navuser.component.ts#L2) (which is ID).  So shouldn't need to hit CMS at all.

src/app/ngrx/reducers/episode.reducer.ts
```diff
@@ -69,7 +69,11 @@ export function reducer(
       };
     }
     case ActionTypes.CASTLE_EPISODE_PAGE_FAILURE: {
-      return {...state, error: action.payload.error};
+      return {
+        ...state,
+        error: action.payload.error,
```
cavis: Ah, so does this mean you're passing remote errors around in the application state, and reacting to that to show some UI?

src/app/ngrx/reducers/selectors/error.selectors.ts
```diff
@@ -1,20 +1,83 @@
...
-    errors.push(`${errorType(episodeError.code)} error occurred while requesting list of episodes`);
-  }
-  return errors;
-});
-export const selectMetricsErrors = createSelector(selectPodcastMetricsError, selectEpisodeMetricsError,
-  (podcasts, episodes) => podcasts.concat(episodes));
-export const selectErrors = createSelector(selectCatalogErrors, selectMetricsErrors, (catalog, metrics) => catalog.concat(metrics));
+export const select500ErrorReloadActions =
+  createSelector(selectRouter,
+    selectPodcastError,
```
cavis: Like the explicit error handlers.  What happens with uncaught global exceptions?  Or do those always land in one of these categories?

src/app/ngrx/reducers/selectors/error.selectors.ts
```diff
@@ -1,20 +1,83 @@
...
+          id: podcastId,
+          group,
+          interval,
+          beginDate,
+          endDate
+        }));
+      }
+      if (podcastTotalsError) {
+        const { podcastId, group, interval, beginDate, endDate } = routerParams;
+        actions.push(new ACTIONS.CastlePodcastTotalsLoadAction({
```
cavis: Yeah, these will probably cover 99% of the castle-type timeout errors out there.

src/app/shared/error/error-retry.component.ts
```diff
@@ -0,0 +1,23 @@
+import { Component, Input } from '@angular/core';
+import { Store } from '@ngrx/store';
+import { AllActions, CmsAccountRetryActionsAction } from '../../ngrx/actions/';
+
+@Component({
+  selector: 'metrics-error-retry',
+  template: `
+    <section *ngIf="retryActions?.length">
+      <p class="error">Something went wrong with loading metrics data. Click 'Retry' to try again.</p>
+      <button class="button" (click)="retry()">Retry</button>
```
cavis: Probably out of scope for this ... but would be nice to somehow capture global uncaught errors, somehow report them to us, and have retry() actually refresh the page.

src/app/shared/table/totals-table.component.ts
```diff
@@ -0,0 +1,61 @@
...
+            {{data.label}}
+          </prx-checkbox>
+        </div>
+        <div class="number">{{data.value | largeNumber}} <span class="percent">({{data.percent.toPrecision(2)}}%)</span></div>
+        <div class="number growth"></div>
+      </div>
+    </div>
+
+    <div class="table">
+      <div *ngIf="tableData?.length > numRowsWithToggle && numRowsWithToggle > 0" class="other header row">
```
cavis: image

I find the start of the "Others:" section a bit weird.  Wish there was some other kind of separator, maybe even an `<hr>` or something.  The white box makes the previous data row look related to "Others" somehow.  But I can be convinced to live with this for now!

src/app/ngrx/actions/cms.action.creator.ts
sandikbarr: It does the userinfo request from the styleguide as well. That gets the username for the dropdown menu. But we also need the user image, which I am pretty sure I can only get from CMS. Honestly, I just left the CMS stuff alone when I pulled out the podcast and episode requests to use castle instead.

cavis: Ah, okay, so remembering more now.  The Publish AppComponent was updated to the user info service instead of CMS.  And that shared user-info-service can actually fetch the userimage itself.  (ID provides [a link](https://id.prx.org/userinfo?scope=profile%20apps) directly to your CMS public user, which has an image in it).

I think Metrics just got missed during that auth cleanup round.

Here it is -> https://github.com/PRX/publish.prx.org/pull/507

sandikbarr:
So are we thinking this needs a change to the retry/error handling? Should it just assume the token is good after a 500 and re-issue the requests?

src/app/ngrx/actions/cms.action.creator.ts
sandikbarr: ah... can we make that a separate ticket?

src/app/ngrx/effects/cms.effects.ts
cavis: Yeah, I guess I would prefer to fix it upstream in the styleguide, so it'll be fixed in Publish too.

It does seem like this does a lot of the same logic as the HalDoc, though at a higher level.  "If you get a 401, re-GET the token from ID, and retry that request once".  Maybe we should open a ticket to test this out in the styleguide?  Should be able to recreate pretty easily at a low level, by adding a char to the token or something to invalidate it.

src/app/ngrx/reducers/selectors/error.selectors.ts
cavis: +1 gotcha

src/app/shared/error/error-retry.component.ts
cavis: Ah, okay, i missed/forgot about that!  "wouldn't-it-be-nice" withdrawn.

Issue #305, opened 2018-09-11
Element collision in header on small screens


farski:
![image](https://user-images.githubusercontent.com/3483/45368046-a2807800-b5b0-11e8-865e-5be57a08fb33.png)

PR #316, opened 2018-09-25
adds country geochart

sandikbarr:
* adds the geochart for countries
* support for mobile and window resizing layouts by redrawing geochart
* podcast dropdown menu:
  * max-height set according to viewport so it does not fall behind zendesk widget
  * max-width with text-overflow to prevent collision of elements in header #305 
* fix for nested geo data to be reloaded when podcast, begin/end dates, or interval changes (ranks only on interval change) (oops, that was a miss on the last release 😬)
* extends entity key/id for nested subdiv data to include the begin and end dates to correctly handle "partial" 500 error handling with nested data


src/app/core/routing/routing.service.ts
```diff
@@ -93,8 +93,15 @@ export class RoutingService {
           this.isBeginDateChanged(newRouterParams) || this.isEndDateChanged(newRouterParams) || this.isIntervalChanged(newRouterParams)) {
           this.loadPodcastRanks(newRouterParams);
         }
-        if (newRouterParams.group === GROUPTYPE_GEOCOUNTRY && this.isFilterChanged(newRouterParams)) {
+        if (newRouterParams.group === GROUPTYPE_GEOCOUNTRY && newRouterParams.filter &&
```
pstjean: A bit tangential to this PR, but does it make sense to set `newRouterParams` as a class property? It seems to be passed to almost every method called in the class.

src/app/ngrx/reducers/models/podcast-totals.model.ts
```diff
@@ -1,11 +1,22 @@
 import { Rank } from './rank.model';
+import { GroupType, GROUPTYPE_GEOSUBDIV } from './group.type';
+import * as dateFormat from '../../../shared/util/date/date.format';
```
pstjean: Not sure if this applies to our version of TypeScript, but it looks like you can now use a `baseUrl` configuration to optionally use absolute imports, which would be handy right here: http://www.typescriptlang.org/docs/handbook/module-resolution.html#base-url

src/app/core/routing/routing.service.ts
sandikbarr: I can see where you're coming from there. this.routerParams is the equivalent, but the newRouterParams is passed around to compare what to do with it by comparing it to the most recent this.routerParams, so it's more like a decision of what to do with the event then it's being saved on the class after the decision is made.

src/app/ngrx/reducers/models/podcast-totals.model.ts
sandikbarr: 😎

Issue #98, opened 2017-11-29
Geolocation chart by state/province/subdivision


sandikbarr:
Show geo chart to view downloads grouped by state, province, or other subdivision. Almost all countries have subdivisions (do we call them subdivisions or states and provinces where applicable?) 

* Only a single country broken to subdivisions can be shown at a time (i.e. no combined british isles or combining the US with Canada)
* unpaged list of all subdivisions within a country
* ~subdivisions can be toggled on/off display~
* ~can show either the full show or filter to a single episode~ initially just the podcast data
* ~the map view can be flipped to a chart view breaking down the totals from the map display into daily, weekly, or monthly intervals for the date range~ #314

sandikbarr:
Grooming TBD when we have more of an opportunity to discuss geolocation metrics.

Issue #94, opened 2017-11-28
Geolocation chart by Nielsen DMA Metro region


sandikbarr:
Shows geo chart to view downloads grouped by Neilsen DMA Metro regions. A DMA region is a group of counties that form an exclusive geographic area in which the home market television stations hold a dominance of total hours viewed. There are 210 DMA regions, covering the entire continental United States, Hawaii, and parts of Alaska.

* US only ~(so US can be drilled down to state or DMA)~
* ~Do DMAs cross state boundaries (they are defined as a group of counties, but it looked as though some may cross state lines)? If not crossing state lines, the US could allow drill down by state then DMA.~ --> Yes, they can cross state lines.
* ~Metro codes are in max mind data but do these regions have names?~ --> Metro names are provided by the API.
* DMAs can be toggled on/off display
* ~can show either the full show or filter to a single episode~ (single episode selection will be implemented in a separate ticket)
* ~the map view can be flipped to a chart view breaking down the totals from the map display into daily, weekly, or monthly intervals for the date range.~ #314 asks which charts views do we want horizontal bar with totals, multi-line with intervals, and stacked with intervals?
* Do we need to license this data? --> #315

sandikbarr:
Grooming TBD when we have more of an opportunity to discuss geolocation metrics.

PR #322, opened 2018-09-28
metro and subdiv geocharts

sandikbarr:
#94 and #98


src/app/geo/geochart.map.component.ts
```diff
@@ -49,19 +52,30 @@ export class GeochartMapComponent implements OnInit, OnChanges, AfterViewInit {
...
-      } else {
-        height = 320;
-      }
+    if (GeochartMapComponent.googleLoaded && google.visualization) {
       const options = {
         displayMode: 'region',
         colorAxis: {colors: this.colors},
         legend: 'none'
       };
+      const { width, height } = this.getMapWidthOrHeight();
```
pstjean: This logic and the following two `if` statements could be combined:
`{width: options['width'], height: options['height']} = this.getMapWidthOrHeight();`

pstjean: ... or `Object.assign(options, this.getMapWidthOrHeight())`

Debatable whether either of those is an improvement.

src/app/shared/table/nested-totals-table.component.ts
```diff
@@ -56,21 +56,11 @@ export class NestedTotalsTableComponent {
   @Input() nestedData: TotalsTableRow[];
   @Input() nestedDataLoading: boolean;
   @Input() nestedDataLoaded: boolean;
-  @Input() nestedDataError: any;
+  @Input() nestedDataErrorActions: ACTIONS.AllActions[];
   @Input() routerParams: RouterParams;
   @Output() discloseNestedData = new EventEmitter<string>();
   getGroupName = getGroupName;
 
-  get nestedDataRetryAction() {
```
pstjean: Looks like this moved into `selectNested500ErrorReloadActions` in the error selector? Following the chain of assignment to get there took me a little while.

src/app/geo/geochart.map.component.ts
sandikbarr: good call 👍

src/app/shared/table/nested-totals-table.component.ts
sandikbarr: Yes! Sorry! I factored this out. I was thinking about adding a retry button in the map component as well, but it didn't actually feel right when I did it.

PR #332, opened 2018-10-18
adds charts for geo interval data

sandikbarr:


src/app/ngrx/reducers/selectors/podcast-ranks.selectors.ts
```diff
@@ -141,13 +145,32 @@ export const selectRoutedPodcastRanksChartMetrics = createSelector(
...
+            return groupsCharted.find(g => g.charted && g.groupName === entry.label &&
+              ((groupRoute !== GROUPTYPE_GEOCOUNTRY &&
+                groupRoute !== GROUPTYPE_GEOMETRO) || entry.label !== 'Other'));
           });
       }
     }
   }
 );
 
+export const selectNestedPodcastRanksChartMetrics = createSelector(
```
pstjean: It seems like this might be a good candidate for a test?

PR #355, opened 2019-01-02
Rename PodcastMetrics to PodcastDownloads

pstjean:


src/app/downloads/downloads-chart-container.component.spec.ts
```diff
@@ -53,6 +53,8 @@ describe('DownloadsChartContainerComponent', () => {
         metricsPropertyName, metrics: ep1Downloads}));
       store.dispatch(new ACTIONS.CastlePodcastMetricsSuccessAction({
         id: podcast.id, metricsPropertyName, metrics: podDownloads}));
+      store.dispatch(new ACTIONS.ChartTogglePodcastAction({
```
pstjean: This test was failing for me since the podcast was not charted, so I added this toggle action. It could be that this is a smell indicating an incompatibility with the non-entity implementation.

src/app/ngrx/reducers/selectors/downloads-chart.selectors.spec.ts
```diff
@@ -42,6 +42,8 @@ describe('Downloads Chart Selectors', () => {
         metricsPropertyName, metrics: ep1Downloads}));
       store.dispatch(new ACTIONS.CastlePodcastMetricsSuccessAction({
         id: podcast.id, metricsPropertyName, metrics: podDownloads}));
+      store.dispatch(new ACTIONS.ChartTogglePodcastAction({
```
pstjean: This test was failing for me since the podcast was not charted, so I added this toggle action. It could be that this is a smell indicating an incompatibility with the non-entity implementation.

src/app/ngrx/reducers/podcast-downloads.reducer.ts
```diff
@@ -0,0 +1,77 @@
...
+          loaded: false,
+          ...(charted && { charted }) // Conditionally add charted as object member
+        }
+      }
+      return adapter.upsertOne(podcastUpdate, state);
+    }
+    case ActionTypes.CASTLE_PODCAST_DOWNLOADS_SUCCESS: {
+      const { id, metricsPropertyName, metrics } = action.payload;
+      const currentIds = <string[]>selectPodcastDownloadsIds(state);
+      const charted = includes(currentIds, id);
```
sandikbarr: One thing I see that's changed is that if you toggle off the podcast from the stacked chart, then change the date range or page, the podcast becomes charted again. In production, the podcast chart is default toggled on when first loaded but stays toggled off if you change the parameters. Here it's charting it by default always util toggled off.

src/app/ngrx/reducers/selectors/downloads-chart.selectors.spec.ts
sandikbarr: I see what's going on here... It's related to the change to the `charted` toggle. The first time the entity is saved to state, it doesn't get the `charted` property because of the includes logic. When the application has a LOAD then a  SUCCESS action, `charted: true` is added. For the tests, we're just cutting right to the SUCCESS action, so it's not getting `charted`.

src/app/ngrx/reducers/podcast-downloads.reducer.ts
pstjean: D'oh! Accidentally flipped that logic. Pushed a fix.

Issue #345, opened 2018-12-07
Google Analytics on New Features in Metrics


IcusFMagni:
* multiple episode select
* episode drop date table
* click in email?

Issue #330, opened 2018-10-16
episode selection menu


sandikbarr:
(minus search)

~The podcast navigation menu becomes both a podcast and episode selection menu.~ There is an episode selection menu. The user can select one or more episodes. When a podcast is selected (clicked? hovered?) a list of the podcast's episodes will be shown with their titles and pub dates. Initially, just the first page of X episodes is loaded, but as the user scrolls the list, further pages are loaded and displayed.

Implementation: need to load episode list as user scrolls

Issue #324, opened 2018-09-28
show user agent or geo data on single episode


sandikbarr:
users should be able to select a single episode to view user agent or geo data

PR #362, opened 2019-01-16
episode selection menu

sandikbarr:
This PR satisfies requirements in #330, #331, and #345 to create an episode selection menu with search and multiselect capabilities. The episode selection menu doesn't yet do anything in the app until #324, so the episode data is loaded and processed into a store but the component is not shown anywhere.


src/app/shared/episode-select/episode-select-dropdown.component.ts
```diff
@@ -0,0 +1,109 @@
...
+          scrollTop + clientHeight >= (scrollHeight - 100) &&
+          this.lastPage + 1 <= this.maxPages) {
+        this.loadEpisodes(this.lastPage + 1, this.searchTerm);
+        this.store.dispatch(new ACTIONS.GoogleAnalyticsEventAction({gaAction: 'episode-select-page-load', value: this.lastPage + 1}));
+      }
+    });
+  }
+
+  loadEpisodesOnSearch(search: string) {
+    this.loadEpisodes(1, search);
```
pstjean: Does dispatching 2 separate actions make the most sense here? My gut says to roll them up into 1 but when I think about it I can't see any downside to doing it this way.

src/app/ngrx/reducers/selectors/error.selectors.ts
```diff
@@ -39,7 +39,8 @@ export const select500ErrorReloadActions =
     }
     if (routerParams.metricsType === METRICSTYPE_DOWNLOADS) {
       if (episodeError && episodeError.status === 500) {
-        actions.push(new ACTIONS.CastleEpisodePageLoadAction({podcastId: routerParams.podcastId, page: 1, all: true}));
+        actions.push(new ACTIONS.CastleEpisodePageLoadAction({
```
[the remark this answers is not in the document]
sandikbarr: Yes! Most often hits a different instance and returns the data on the retry. We were having a lot of issues with out of memory on Castle, so this became kinda useful at one point.

src/app/shared/episode-select/episode-select-dropdown.component.ts
sandikbarr: I suppose the google analytics could be handled as a (side) effect when CastleEpisodeSelectPageLoadAction is dispatched, have an effect to dispatch the GoogleAnalyticsEventAction (which is itself an effect actually, `fromGAEvent$`) but it feels more direct to me to just dispatch both actions at their source. Mike Ryan did a great talk at ng-conf last year about designing Actions if you wanna learn more about their thoughts: https://www.youtube.com/watch?v=JmnsEvoy-gY

pstjean: Yeah, thinking about it more I think that's correct. Moving it to the side effect level means that it's a layer of abstraction away from the user interaction, which is really what you're tracking. I'll check out that talk.

PR #367, opened 2019-01-29
aggregate episode data

sandikbarr:


src/app/ngrx/reducers/selectors/episode-ranks.selectors.ts
```diff
@@ -0,0 +1,150 @@
...
+      const rows = aggregateIntervals(episodeRanks, groupsCharted);
+      return rows && rows.filter((row: TimeseriesChartModel) => row.label !== 'Other' || getTotal(row.data) !== 0);
+    }
+  }
+);
+
+export const selectNestedEpisodesRanksChartMetrics = createSelector(
+  selectNestedEpisodesRanks,
+  (episodeRanks: EpisodeRanks[]): TimeseriesChartModel[] => {
+    // TODO: filter Other
```
sandikbarr: finished, remove TODO

src/app/shared/util/chart.util.ts
```diff
@@ -145,3 +150,159 @@ export const minY = (chartType: ChartType): number => {
     return 0;
   }
 };
+
+export const aggregateTotalDownloads = (episodeTotals: EpisodeTotals[]) => {
+  return episodeTotals && episodeTotals.reduce((accTotal, et: EpisodeTotals) => {
```
pstjean: I did not know you could do this! Chain `&&` and get the final value if all evaluate to `true`. Nifty.

pstjean:
This looks good! We talked a little about the complexity of the router in general. I think keeping that as a watch to make sure it doesn't become unmanageably huge is a good idea, although I don't have much in the way of helpful suggestions to improve it. The one thing I did see is a lot of repeated sub-expressions within the conditionals that could be candidates for refactoring into a function, although that would have to improve readability in order to be worth doing.

src/app/shared/util/chart.util.ts
sandikbarr: These kind of null safety checks are definitely something that I've seen our Angular apps error on, so I try to be careful about them.

sandikbarr:
yeah I think looking at refactoring parts of the router service could be helpful as it has grown and grown. I'm often just changing a small part and trying to minimize the changes since that's so central to the application.

PR #371, opened 2019-02-13
CSV exports

sandikbarr:


src/app/ngrx/reducers/group-charted.reducer.ts
```diff
@@ -27,22 +27,6 @@ export function reducer(
       }, state);
     }
 
-    // when ranks are loaded, add them to state
```
sandikbarr: I changed this reducer to only store the groups that have been toggled off/on by the user and selectors that use this store now assume that groups are charted implicitly unless set to false here in this store. This is related to the `isGroupCharted` function you'll see elsewhere.

src/app/ngrx/reducers/selectors/downloads-chart.selectors.spec.ts
```diff
@@ -78,6 +78,13 @@ describe('Downloads Chart Selectors', () => {
     it('should sort episode chart data by total biggest to smallest', () => {
       expect(getTotal(result[0].data)).toBeGreaterThanOrEqual(getTotal(result[1].data));
     });
+
+    it('should only show podcast data if not the same number of episode downloads datapoints', () => {
```
sandikbarr: Danika found a fun bug. Due to the way the reducers are just storing the array of data without regard for changing date ranges or other parameters, if you load the app with one date range, navigate around the pages, then load an extended date range, and navigate around the pages again, an error occurs in the loading stage when the podcast downloads array doesn't line up with the episodes.

src/app/ngrx/reducers/selectors/downloads-chart.selectors.ts
```diff
@@ -68,16 +87,16 @@ export const selectDownloadChartMetrics = createSelector(
     switch (routerParams.chartType) {
       case CHARTTYPE_STACKED:
         if (chartedPodcastDownloads && PodcastDownloads.charted &&
-          chartedEpisodeDownloads && chartedEpisodeDownloads.length) {
-          // if we have episodes to combine with podcast total
+          chartedEpisodeDownloads && chartedEpisodeDownloads.length &&
+          chartedEpisodeDownloads.every(d => chartedPodcastDownloads.data.length === d.data.length)) {
```
sandikbarr: This fixes the bug Danika found to check for matching downloads arrays

src/app/ngrx/reducers/selectors/export.selectors.ts
```diff
@@ -0,0 +1,220 @@
...
+    case CHARTTYPE_EPISODES:
+      if (chartedEpisodeDownloads && chartedEpisodeDownloads.length) {
+        exportData = chartedEpisodeDownloads;
+      }
+      break;
+  }
+  return exportData;
+});
+
+export const toCsvArray = (downloads: ExportData[], dateFormat: Function): string[][] => {
```
pstjean: Munge!

PR #376, opened 2019-02-27
ng6 upgrade

sandikbarr:


src/app/app.component.css
```diff
@@ -1,6 +1,9 @@
 prx-header {
   max-width: 100%;
 }
+prx-header >>> nav {
```
sandikbarr: slight regression on the recent changes to prx-header. I needed to push this dropdown menu back over to the left with a wider parent element.

src/app/ngrx/reducers/episode-alltime-downloads.reducer.ts
```diff
@@ -18,36 +26,30 @@ export function reducer(
 
     case ActionTypes.CASTLE_EPISODE_ALLTIME_DOWNLOADS_LOAD: {
       const { guid, podcastId } = action.payload;
-      return {
-        ...adapter.upsertOne({
+      return adapter.upsertOne(
```
sandikbarr: The ngrx/entity upsert interface changed to no longer use that changes business and just take the whole entity record.

src/app/ngrx/reducers/episode-alltime-downloads.reducer.ts
```diff
@@ -4,12 +4,20 @@ import { AllActions, ActionTypes } from '../actions';
 
 export type State = EntityState<EpisodeAllTimeDownloads>;
 
-export const adapter: EntityAdapter<EpisodeAllTimeDownloads> = createEntityAdapter<EpisodeAllTimeDownloads>({
-  selectId: (e: EpisodeAllTimeDownloads) => e.guid,
```
sandikbarr: I had some changes to make to ngrx for the upgrade, so I also stopped using this selectId entityAdapter helper. I feel like it wasn't actually doing me any favors. The `id` property was still being added the the entity records under the hood. It would show up in marble tests and you'd have to account for it being there when it wasn't on the interface to the entity record. It just wasn't helping as much as I thought it would.

src/app/ngrx/reducers/episode-downloads.reducer.ts
```diff
@@ -29,32 +29,53 @@ export function reducer(state: State = initialState, action: AllActions) {
     case ActionTypes.CASTLE_EPISODE_DOWNLOADS_LOAD: {
       const { guid, podcastId, page } = action.payload;
       return adapter.upsertOne(
-        {id: guid, changes: { guid, podcastId, page, error: null, loading: true, loaded: false}},
+        {
+          id: guid,
+          ...selectEpisodeDownloadsEntities(state)[guid],
```
sandikbarr: Since upsert doesn't take `changes` anymore, I tend to just prefill it with the current record when I'm not intending to set all the values with an action.

src/app/ngrx/reducers/episode-downloads.reducer.ts
```diff
@@ -29,32 +29,53 @@ export function reducer(state: State = initialState, action: AllActions) {
...
+          ({
+            id: guid,
+            guid,
+            ...selectEpisodeDownloadsEntities(state)[guid],
+            charted: guid === action.payload.guid})),
           state);
     }
     case ActionTypes.CHART_TOGGLE_EPISODE: {
       const { guid, charted } = action.payload;
       return adapter.upsertOne(
```
sandikbarr: TODO: this could just be an update rather than an upsert. Toggles are meant to only change existing records.

src/app/ngrx/reducers/episode-ranks.reducer.spec.ts
```diff
@@ -6,7 +6,7 @@ import {
   ep0AgentNameRanks,
   ep0AgentNameDownloads
 } from '../../../testing/downloads.fixtures';
-import { GroupType, GROUPTYPE_AGENTNAME, METRICSTYPE_TRAFFICSOURCES, episodeRanksKey } from './models';
+import { GroupType, GROUPTYPE_AGENTNAME, METRICSTYPE_TRAFFICSOURCES, episodeRanksId } from './models';
```
sandikbarr: This is a naming change. I just started using the id field instead of adding a separate combo-key and pointing to it with the `selectId` helper.

src/app/ngrx/reducers/episode-alltime-downloads.reducer.ts
```diff
@@ -4,12 +4,20 @@ import { AllActions, ActionTypes } from '../actions';
...
-});
+export const adapter: EntityAdapter<EpisodeAllTimeDownloads> = createEntityAdapter<EpisodeAllTimeDownloads>();
 
 export const initialState: State = adapter.getInitialState({});
 
+export const {
```
sandikbarr: You'll see that I moved these declarations above the reducer in a lot of places where I wanted to use them inside the reducer to select existing entity records.

src/app/ngrx/reducers/episode-totals.reducer.spec.ts
```diff
@@ -23,23 +23,22 @@ describe('Episode Totals Reducer', () => {
...
     const { group, filter, beginDate, endDate } = routerParams;
-    const key = episodeTotalsKey(guid, group, filter, beginDate, endDate);
+    const id = episodeTotalsId(guid, group, filter, beginDate, endDate);
     const newState = reducer(initialState,
       new ACTIONS.CastleEpisodeTotalsSuccessAction({
         guid, group, beginDate, endDate,
         ranks: ep0AgentNameRanks}));
-    expect(newState.entities[key]).toEqual({
+    expect(newState.entities[id]).toEqual({
       // @ts-ignore using upsert adds 'id' property to entity, seems like ngrx/entity v6 gets rid of this
```
sandikbarr: TODO: remove stale comment

src/app/ngrx/reducers/podcast-downloads.reducer.ts
```diff
@@ -21,45 +20,32 @@ export function PodcastDownloadsReducer(state: PodcastDownloadsState = initialSt
...
-        id,
-        changes: {
-          error: null,
-          loading: true,
-          loaded: false,
-          ...(charted && { charted }) // Conditionally add charted as object member
-        }
-      };
-      return adapter.upsertOne(podcastUpdate, state);
+      const entity = selectPodcastDownloadsEntities(state)[id];
```
sandikbarr: I temp broke this again in making these upgrade changes, so to review here is how `charted` should work: if the podcast entity has not been added to the state before, it should be charted. It is only false if the user has toggled it off, so the entity would already exist on state and be false.

src/app/ngrx/reducers/podcast-totals.reducer.ts
```diff
@@ -1,59 +1,59 @@
...
   state = initialState,
   action: AllActions
 ): State {
   switch (action.type) {
     case ActionTypes.CASTLE_PODCAST_TOTALS_LOAD: {
       const { id, group, filter, beginDate, endDate } = action.payload;
-      const key = podcastTotalsKey(id, group, filter, beginDate, endDate);
-      return {
-        ...adapter.upsertOne({
+      const key = podcastTotalsId(id, group, filter, beginDate, endDate);
```
sandikbarr: TODO: I wanted to sort out this key vs id thing. Use podcastId (not id) when it is a combo key.

angular.json
```diff
@@ -0,0 +1,143 @@
...
+                "glob": "*",
+                "input": "node_modules/ngx-prx-styleguide/assets/images",
+                "output": "assets/images/"
+              }
+            ],
+            "styles": [
+              "node_modules/c3/c3.css",
+              "node_modules/pikaday/css/pikaday.css",
+              "node_modules/pikaday/css/triangle.css",
+              "node_modules/ngx-prx-styleguide/assets/styles/bundle.scss",
```
pstjean: You'll need the `ng-select` stylesheet here if you ever wanted to use that element (or the new tagging element Rick's building) from Styleguide.

src/app/ngrx/effects/castle.effects.ts
```diff
@@ -287,24 +281,24 @@ export class CastleEffects {
     ofType(ACTIONS.ActionTypes.CASTLE_PODCAST_RANKS_LOAD),
     map((action: ACTIONS.CastlePodcastRanksLoadAction) => action.payload),
     mergeMap((payload: ACTIONS.CastlePodcastRanksLoadPayload) => {
-      const { id, interval, group, filter, beginDate, endDate } = payload;
+      const { podcastId, interval, group, filter: groupFilter, beginDate, endDate } = payload;
```
pstjean: I love how destructuring renaming makes a rename for clarity like this so easy.

src/app/ngrx/reducers/episode-alltime-downloads.reducer.ts
pstjean: Speaking of destructuring renaming, I like using it on the results of `getSelectors()` to simplify things a bit here. It looks like we're exporting both the original and aliased names for these particular selectors too, which might lead to some confusion in the future (obviously if both are used and we need to preserve this for now that's a different story):

```
export const {
  selectIds: selectEpisodeAllTimeDownloadsGuids,
  selectEntities: selectEpisodeAllTimeDownloadsEntities,
  selectAll: selectAllEpisodeAllTimeDownloads,
} = adapter.getSelectors();
```

tsconfig.json
```diff
@@ -14,6 +14,8 @@
     "lib": [
       "es2016",
       "dom"
-    ]
+    ],
+    "module": "es2015",
+    "baseUrl": "./"
```
pstjean: One of the steps on the upgrade checklist was to tell the Angular compiler specifically to stop preserving whitespace: PRX/publish.prx.org@193d8a6dd80

It's not really made clear why, but I made the change and just assumed it was added to the list for a good reason.

angular.json
sandikbarr: long story, that multiselect got moved out of metrics a while ago on purpose and it has its own dropdown/select menus, so _if_ it gets added back in, then I'll add the stylesheet.

tsconfig.json
sandikbarr: sure missed that, thanks!

src/app/ngrx/effects/castle.effects.ts
sandikbarr: `groupFilter` is def more clear naming, but I was honestly doing it because it came up in linting. I was shadowing RxJS filter. Oops! If I had it to do over again, the `RouterParams` `filter` would be called `groupFilter`. You only get so many do overs per PR.

src/app/ngrx/reducers/episode-alltime-downloads.reducer.ts
sandikbarr: Good call. My earlier pattern was just following what I'd learned in beginner tutorials, but I agree, there's good reason to go this route.

PR #378, opened 2019-03-12
support for url to episode downloads

sandikbarr:


src/app/ngrx/effects/routing.effects.ts
```diff
@@ -25,7 +26,12 @@ export class RoutingEffects {
     ofType(ROUTER_NAVIGATION),
     map((action: RouterNavigationAction) => action.payload),
     switchMap((payload: RouterNavigationPayload<any>) => {
-      const routerParams: RouterParams = {...payload.routerState};
+      // drop guids from router params
```
sandikbarr: Initially I was trying to keep the episode urls on the route unless they became unselected from the episode selection dropdown, the chart toggle, or the chart single episode action (clicking on an episode from the podcast chart takes you to the multiline paged downloads with that single episode selected.) That was not a good move. Essentially I was keeping track of two lists, both the selected episodes and the episodes on the router state. A little background on this, early on with metrics it was stated that the url should reflect the full parameter state and we had episode guids in there for a while just on the paged downloads. The url would grow and grow with episode guids, and there was feedback that saving the chart toggle state was not desired. So I dropped it. I can't put all the selected episodes on the route because there is no limit to the number of episodes a user can select through the application, but urls do have a length limit. So long story, short here- what this does is takes in a guids parameter and dispatches it to the selected episode store. Th episode select store saves the state, and router store isn't aware of the selected episodes state.

PR #380, opened 2019-04-09
Fixes google sheets change detection

sandikbarr:


src/app/shared/menu/export/export-google-sheets.service.ts
```diff
@@ -103,6 +104,9 @@ export class ExportGoogleSheetsService {
       })
       .then(() => {
         this._state.next({...this._state.getValue(), sheet: googleSheet, busy: false});
+        // explicitly call change detection (to hide spinner)
+        this.appRef.tick();
```
sandikbarr: The parent dropdown component that contains the spinner wasn't picking up on the fact that export state is no longer busy, so the spinner was still showing. This is like the old $scope.apply() in angularjs. It seems relatively harmless to add this to fix the problem since it only gets called explicitly on a google sheets export, and AppplicationRef.tick is what NgZone uses for change detection.

cavis: +1

PR #399, opened 2019-10-09
Fix dropday export data to be cumulative

sandikbarr:


src/app/ngrx/reducers/selectors/export.selectors.spec.ts
```diff
@@ -34,53 +33,68 @@ describe('Export Selectors', () => {
       dispatchHelper.dispatchEpisodeDownloads(store);
     });
 
-    it('should have download exports', () => {
+    it('should have podcast download exports', done => {
```
sandikbarr: These async tests weren't using `done()`, and turns out some of them weren't actually passing, whoops fixed!

src/app/ngrx/reducers/selectors/export.selectors.spec.ts
```diff
@@ -94,19 +108,13 @@ describe('Export Selectors', () => {
       dispatchHelper.dispatchEpisodeDropday(store);
     });
 
-    it('should have dropday exports', () => {
-      dispatchHelper.dispatchRouterNavigation(store, {chartType: CHARTTYPE_HORIZBAR});
```
sandikbarr: no HORIZBAR export on drop day in app anymore

PR #412, opened 2019-10-14
upgrade to angular 7

sandikbarr:
also cleans up some subscriptions in OnDestroy and some async specs


src/app/ngrx/reducers/selectors/group-charted.selectors.spec.ts
```diff
@@ -21,17 +21,21 @@ describe('Group Charted Selectors', () => {
     dispatchHelper.dispatchRouterNavigation(store, {metricsType: METRICSTYPE_TRAFFICSOURCES, group: GROUPTYPE_AGENTNAME});
   });
 
-  it('should have just the toggled groups charted status', () => {
+  it('should toggled groups charted off', done => {
```
pstjean: "should toggled" `=>` "should toggle"

PR #418, opened 2020-06-09
angular 8 upgrade

### LINKED ISSUE #417: ng8 upgrade

sandikbarr:
Closes #417 

* The Router Serializer was updated to include the required url property in the route params state, but otherwise I didn't change to the newer NgRx API yet in this PR.
* Expression changed after checked on date range picker is unresolved and tracked down a weird emit from the underlying pikaday date picker library. This type of error is only ever thrown in development mode.


src/app/ngrx/reducers/episode-dropday.reducer.spec.ts
```diff
@@ -6,10 +6,15 @@ import { INTERVAL_DAILY } from './models';
 describe('EpisodeDownloadsReducer', () => {
   let newState: fromEpisodeDropday.State;
   beforeEach(() => {
```
sandikbarr, line 8: Handful of type errors came up in the tests

sandikbarr:
I am still seeing an Angular Expression Changed After Checked error, which only gets thrown in development mode. I tracked it down to the Pikaday datepicker library which seems to be no longer maintained. I can get it the error to happen pretty consistently if I load the app with the default date range then select "Last year JAN 1  - DEC 31" from the date options. During the date change, for some reason pikaday emits a date from May 2020 for the end date (the month prior to what was selected) that bubbles up into a date range change on the metrics side that evaluates to an "Other" date range (doesnt match one of the given ranges) which makes another change happen programmatically during the change cycle and throws this error.

PR #428, opened 2020-10-12
angular 9 upgrade

sandikbarr:
closes #424

PR #435, opened 2021-02-04
Fix Drop-Day tab days-picker

### LINKED ISSUE #431: Drop Day range picker not working
On the Drop Day tab:

- the range on the graph (10 days) doesn't reflect what's selected by default (28 days)
- changing the days does nothing, or gives you a "cannot read property 'type' of null" error

image

cavis:
Closes #431.

Looks like back in #428 (and b0a69eba09755d063b35841000f452a01aa669eb) ngrx was upgraded from 8.6 -> 9.2.  Which [deprecated the @Effect decorator](https://ngrx.io/guide/migration/v9#ngrxeffects).

One decorator was missed apparently (only one in the codebase).  And using both the decorator `createEffect` does weird things - was complaining about the null returned by the effect having no `type`?  (I think?).

In any case, removing the decorator fixes it.

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -6,10 +6,16 @@ env:
     PRX_ECR_CONFIG_PARAMETERS: "MetricsEcrImageTag"
   parameter-store:
     CODECOV_TOKEN: "/prx/test/metrics.prx.org/CODECOV_TOKEN"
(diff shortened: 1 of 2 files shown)


svevang (approved):
👍

PR #441, opened 2022-05-17
Modernize docs and dev setup

farski:
Could be very wrong.


.nvmrc
```diff
@@ -1 +0,0 @@
-10.15.3
```
cavis, line 1: Does it hurt to keep `.nvmrc` around?  We still have many node projects that only include this file, and `asdf` is compatible with it.

Makefile
```diff
@@ -20,4 +20,14 @@ lint:
 
 check: lint test
 
-.PHONY: run build test clean install all check lint
+dev-init:
+	asdf install
+	npm install -g npm@latest
+	npm install -g yarn@latest
+	yarn install
+	echo 4202 > ~/.puma-dev/metrics.prx
```
cavis, line 28: I feel like we've moved away from Makefiles at this point.  And these minimal commands could be included in the docs.  But then again, you're keeping things in line with the current codebase.

README.md
```diff
@@ -3,95 +3,20 @@ PRX Metrics Frontend App
 
 This app is used to view podcast metrics.
 
-# Install
+## Local Development
 
-## API and Backend Dependencies
+See [CONTRIBUTING](/PRX/metrics.prx.org/blob/master/CONTRIBUTING.md) for information on installing and configuring this project's dependencies, such as [ID](https://github.com/PRX/id.prx.org/) and [CMS](https://github.com/PRX/cms.prx.org/).
```
cavis, line 8: Honestly, seems a bit weird to me to have a 20 line README, and a 55 line CONTRIBUTING.  When the info is that short, I'd prefer a single file with a `# Contributing` section.

CONTRIBUTING.md
```diff
@@ -0,0 +1,54 @@
+
+## API and Backend Dependencies
```
cavis, line 2: I guess these aren't the sort of things I'd normally expect in a CONTRIBUTING doc?  The way Github links to them, they're normally used to communicate rules and processes for contributing to an OSS project.  Not information on installing/setup/running-tests.

cavis:
@farski do you want to keep this open?

farski:
Not if you're going to keep making fun of my Makefiles. I think we have a TODO somewhere to have a team discussion about project standards, so we can probably revisit this once we have some outcome there. Also I guess Metrics may disappear at some point, so maybe not ever worth it.

PR #445, opened 2022-08-30
Switch to AGPL-3.0

farski:


cavis:
Also needs a `package.json` change.

cavis (approved):
👍

PR #449, opened 2025-04-10
WIP: Global Nav scaffolding

### LINKED ISSUE #448: Apply Global Nav
We should bite the bullet and update for smooth navigation between apps. 

Impetus: received feedback from multiple folks about this and the Metrics replacement isn't close enough.

brandonhundt:
closes #448 

This gets us close enough? 

Still needs to be wired up.


src/app/app.component.css
```diff
@@ -1,3 +1,180 @@
...
+  position: absolute !important;
+  width: 1px !important;
+  height: 1px !important;
+  padding: 0 !important;
+  margin: -1px !important;
+  overflow: hidden !important;
+  clip: rect(0, 0, 0, 0) !important;
+  white-space: nowrap !important;
+  border: 0 !important;
+}
```
brandonhundt, line 11: There might be an existing class that does this already. We can use that. This is classic Bootstrap.

src/app/app.component.html
```diff
@@ -1,11 +1,52 @@
...
+    <a class="prx-user-menu-button" href="//id.prx.org/profile">Manage Account</a>
+    <a class="prx-user-menu-item" href="/sessions/refresh">Reload</a>
+    <a class="prx-user-menu-item" href="/sessions/logout">Logout</a>
+  </div>
+</nav>
+</div>
+
+<div class="prx-app-bar">
+  <metrics-nav-menu *ngIf="authorized$ | async"></metrics-nav-menu>
+</div>
```
brandonhundt, line 48: I wasn't able to test this, but the idea here is that if the user is logged in, they should get the podcast switcher in the app nav bar (like Feeder's podcast switcher). If I can figure out how to log in to the app, I can ensure the styles work for it.

src/app/app.component.html
```diff
@@ -14,5 +55,4 @@
     </div>
   </article>
 </main>
-<prx-footer></prx-footer>
```
brandonhundt, line 17: No longer needed with the global nav. However, removing the footer leaves a large void at the end of the page. Again, if I can log in, I can take a look.

src/app/app.component.html
```diff
@@ -1,11 +1,52 @@
 <prx-auth [host]="authHost" [client]="authClient"></prx-auth>
-<prx-header>
-  <metrics-podcast-nav *ngIf="loggedIn$ | async"></metrics-podcast-nav>
-  <prx-navuser *ngIf="loggedIn$ | async" [userinfo]="userinfo$ | async">
-    <prx-spinner class="user-loading"></prx-spinner>
-    <prx-image *ngIf="userDoc$ | async" class="user-loaded" [imageDoc]="userDoc$ | async"></prx-image>
-  </prx-navuser>
-</prx-header>
```
brandonhundt, line 8: I chose not to fight style guide.

cavis (approved):
👍

PR #451, opened 2026-07-20
Exec app process from bin/application

svevang:
Multi-repo changeset

bin/application is the container ENTRYPOINT and starts the app as a child process without exec, so the app never receives the SIGTERM that ECS sends on task stop. Where the task definition sets InitProcessEnabled, docker-init forwards SIGTERM only to the wrapper shell, and the app is SIGKILLed when the shell exits. Where it doesn't, the wrapper shell is PID 1, ignores SIGTERM, and the task waits out the full StopTimeout before SIGKILL. Either way there is no graceful shutdown: workers are killed mid-job and web servers never drain.

This execs the selected app process so it receives SIGTERM directly. Callers that previously supplied exec themselves are updated to remove the now-redundant argument.


bin/application
```diff
@@ -22,12 +22,12 @@ ApplicationRun () {
   elif [ "$PROCESS" = "--" ] ; then
     CMD=
   elif [ "$PROCESS" = "serve" ] ; then
-    CMD="npm run-script serve"
+    CMD="node_modules/.bin/prx-ng-serve dist"
```
cavis, line 25: Why also change the command?  Does `run-script` not relay signals?

cavis (approved):
👍 probably fine

bin/application
svevang: That's right, invoked that way we end up with a process chain that is four deep (including init)

```
/app # pstree
docker-init---application---npm run-script -+-node---10*[{node}]
                                            |-6*[{node}]
                                            `-4*[{npm}]
```

